PR #9, opened 2018-02-05
add tool to check data

WangXiangUSTC:


sync_check/util/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+	}
+	if cnt == 0 {
+		log.Infof("no data found in %s.%s", dbname, tableName)
+		return noChunks, nil
+	}
+
+	chunkCnt := cnt / int64(chunkSize)
+	if sample != 100 {
+		// create more chunk for sampling check
+		chunkCnt *= 10
```
WangXiangUSTC: if sample != 100 means use sampling check, and can check more fragmented by split to more chunk.

sync_check/util/chunk.go
```diff
@@ -0,0 +1,313 @@
...
+			return nil, errors.Trace(err)
+		}
+		exist := false
+		column, exist = GetColumnByName(table, splitField)
+		if !exist {
+			return nil, fmt.Errorf("can't find column %s in table %s", splitField, tableName)
+		}
+	}
+
+	timeRange := "true"
```
WangXiangUSTC: set timeRange to "true" can make the code more simple, no need to judge the timeRange's value.

sync_check/util/diff.go
```diff
@@ -0,0 +1,574 @@
...
+
+	checkNums := len(dumpJobs) * df.sample / 100
+	checkNumArr := getRandomN(len(dumpJobs), checkNums)
+	log.Infof("total has %d check jobs, check %+v", len(dumpJobs), checkNumArr)
+
+	checkResultCh := make(chan bool, df.checkThCount)
+	defer close(checkResultCh)
+
+	for i := 0; i < df.checkThCount; i++ {
+		go df.checkChunkDataEqual(dumpJobs, checkNumArr[checkNums*i/df.checkThCount:checkNums*(i+1)/df.checkThCount], tblName, checkResultCh)
```
WangXiangUSTC: every goroutine check some chunk's data

sync_check/util/diff.go
```diff
@@ -0,0 +1,574 @@
...
+}
+
+func (df *Diff) equalTableData(tblName string) (bool, error) {
+	dumpJobs, err := generateDumpJob(df.db1, df.dbName, tblName, df.timeField, df.beginTime, df.endTime, df.splitField, df.chunkSize, df.sample)
+	if err != nil {
+		return false, errors.Trace(err)
+	}
+
+	checkNums := len(dumpJobs) * df.sample / 100
+	checkNumArr := getRandomN(len(dumpJobs), checkNums)
```
WangXiangUSTC: for example, the len(dumpJobs) is 10, and sample is 50, means only check 50% chunks is ok. 
the total job's id is [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] , and getRandomN may return [9, 1 ,2 , 4, 7]

sync_check/util/chunk.go
holys: It's better to comment it in the code.

sync_check/util/db.go
```diff
@@ -0,0 +1,340 @@
...
+	return fmt.Sprintf("DBConfig(%+v)", *c)
+}
+
+// CreateDB create a mysql fd
+func CreateDB(cfg DBConfig) (*sql.DB, error) {
+	dbName := cfg.Name
+	createDBSql := fmt.Sprintf("create database if not exists %s", dbName)
+	// dont't have database
+	dbDSN := fmt.Sprintf("%s:%s@tcp(%s:%d)/mysql?charset=utf8", cfg.User, cfg.Password, cfg.Host, cfg.Port)
+	db, err := sql.Open("mysql", dbDSN)
```
holys: It seems no need to open db twice.

sync_check/util/chunk.go
holys: Please commit it in the code

holys:
The rest LGTM

sync_check/config.go
```diff
@@ -0,0 +1,116 @@
...
+
+	ConfigFile string
+
+	Delay int `toml:"delay" json:"delay"`
+
+	ChunkSize int `toml:"chunk-size" json:"chunk-size"`
+
+	Sample int `toml:"sample" json:"sample"`
+
+	CheckThCount int `toml:"check-thcount" json:"check-thcount"`
```
silentsai, line 77: what's `Th` for ? Try to name exportable variable in detail

sync_check/config.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// configFromFile loads config from file.
+func (c *Config) configFromFile(path string) error {
+	_, err := toml.DecodeFile(path, c)
+	return errors.Trace(err)
+}
+
+func (c *Config) checkConfig() bool {
+	if c.Sample > 100 || c.Sample < 0 {
```
silentsai: avoid magic number, defined as const instead would be better

sync_check/util/diff.go
```diff
@@ -0,0 +1,577 @@
...
+		checkNums = 1
+	}
+	checkNumArr := getRandomN(len(dumpJobs), checkNums)
+	log.Infof("total has %d check jobs, check %+v", len(dumpJobs), checkNumArr)
+
+	checkResultCh := make(chan bool, df.checkThCount)
+	defer close(checkResultCh)
+
+	for i := 0; i < df.checkThCount; i++ {
+		go df.checkChunkDataEqual(dumpJobs, checkNumArr[checkNums*i/df.checkThCount:checkNums*(i+1)/df.checkThCount], tblName, checkResultCh)
```
silentsai: slice is not thread safe , you can't guarantee that evil things occurs inside the below function ~
so pick up the corresponding job into a new slice and then pass it as the input arg to `checkChunkDataEqual ` would be better , which could also simplify the defenition of func ~

sync_check/util/diff.go
```diff
@@ -0,0 +1,577 @@
...
+				return true, nil
+			}
+		}
+	}
+}
+
+func (df *Diff) checkChunkDataEqual(dumpJobs []*dumpJob, checkNumArr []int, tblName string, checkResultCh chan bool) {
+	log.Infof("check array: %+v", checkNumArr)
+	if len(checkNumArr) == 0 {
+		checkResultCh <- true
```
silentsai: return true/false as the function's returnning  value
fill it into channel outside base on the determination of calling side

sync_check/util/diff.go
```diff
@@ -0,0 +1,577 @@
...
+		return
+	}
+
+	eq := equalStrings(tbls1, tbls2)
+	if !eq {
+		log.Infof("show tables get different table. [source db tables] %v [target db tables] %v", tbls1, tbls2)
+		equal = false
+	}
+
+	for _, tblName := range tbls1 {
```
yanyanqing: the tbls1 may have many table，need compare all ? u can compare already Synchronized table i guess

sync_check/config.go
WangXiangUSTC: th means thread

sync_check/util/diff.go
WangXiangUSTC: add a config, user can choose the table to check

sync_check/config.go
WangXiangUSTC: done

sync_check/util/diff.go
WangXiangUSTC: done

silentsai:
LGTM

PR #12, opened 2018-03-24
support use row_id as split column

WangXiangUSTC:


WangXiangUSTC:
@LinuxGit @holys PTAL

sync_check/config.go
```diff
@@ -77,6 +77,8 @@ type Config struct {
 	CheckThCount int `toml:"check-thcount" json:"check-thcount"`
 
 	Tables []string `toml:"tables" json:"tables"`
+
+	UseRowID bool `toml:"use-rowid" json:"use_rowid"`
```
holys: Should we keep consistent in json tag? `use_rowid` -> `use-rowid `

holys:
rest LGTM

PR #24, opened 2018-06-07
pkg: construct base libraries

IANTHEREAL:
We should build our base library and open it, like the feature that splits a table into chunks, etc.
It's good to reuse these library to develop some small tools.
I'm going to start with this PR and break up some of the libraries of TiDB-Binlog firstly. 

@july2993 @WangXiangUSTC PTAL


IANTHEREAL:
I fix vendor in pr #25

IANTHEREAL:
ping @WangXiangUSTC @july2993

pkg/utils/tso.go
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+import (
+	"time"
+)
+
+// TSOToTime translates tso to timestamp
```
WangXiangUSTC: s/timestamp/time/

WangXiangUSTC:
LGTM

pkg/utils/tso.go
```diff
@@ -0,0 +1,24 @@
...
+
+package utils
+
+import (
+	"time"
+)
+
+// TSOToTime translates tso to time
+func TSOToTime(ts int64) time.Time {
+	t := time.Unix(ts>>18/1000, 0)
```
july2993: what about to be more precise get the Millisecond part to

IANTHEREAL: I think @WangXiangUSTC want a fuzzy time here, I would add comments or change it.  But anther problem, do we need a time zone in `generate savepoint`?

WangXiangUSTC: still need time zone, we can retain the code now, I will refine it next pr.

IANTHEREAL: I refine it later

generate_binlog_position/meta.go
```diff
@@ -94,7 +94,7 @@ func (lm *localMeta) Save(ts int64, timeZone string) error {
 	}
 
 	if timeZone != "" {
-		t := pkg.TsToTime(ts)
+		t := pkg.TsToRoughTime(ts)
```
IANTHEREAL, line 97: I would remove `generate_binlog_position` in next pr, whatever

IANTHEREAL:
@july2993 PTAL

july2993 (approved):
LGTM

PR #34, opened 2018-06-27
*: add simple table structure check

IANTHEREAL:
* add table charset checking
* add table foreign key checking
* modify old checker code


IANTHEREAL:
@csuzhangxc @dantin PTAL

pkg/check/table_structure.go
```diff
@@ -0,0 +1,148 @@
...
+
+	var err error
+
+Loop:
+	for schema, tables := range c.tables {
+		if len(tables) == 0 {
+			tables, err = dbutil.GetTables(ctx, c.db, schema)
+			if err != nil {
+				markCheckError(r, err)
+				break Loop
```
dantin: How about return `Result` here instead of break to outer loop.

pkg/check/table_structure.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+		}
+
+		for _, table := range tables {
+			statement, err := dbutil.GetCreateTableSQL(ctx, c.db, schema, table)
+			if err != nil {
+				markCheckError(r, err)
+				break Loop
```
dantin: ditto

checker/main.go
```diff
@@ -70,23 +53,32 @@ func main() {
...
+
+	schema := flag.Args()[0]
+	tables := flag.Args()[1:]
+	checkTables(schema, tables)
+	log.Infof("complete checking!!")
+}
+
+func checkTables(schema string, tables []string) {
+	db, err := dbutil.CreateDB(dbutil.DBConfig{
+		User:     *username,
```
csuzhangxc: how about rename `dbutil.CreateDB` to `dbutil.OpenDB`, match `dbutil.CloseDB`

IANTHEREAL: yep, it's better

pkg/check/table_structure.go
```diff
@@ -0,0 +1,148 @@
...
+		db:     db,
+		tables: tables,
+	}
+}
+
+// Check implements Checker interface
+func (c *TablesChecker) Check(ctx context.Context) *Result {
+	r := &Result{
+		Name:  c.Name(),
+		Desc:  "table structure compatibility check",
```
csuzhangxc, line 37: "checks table structure compatibility"? similar to other checkers

pkg/check/table_structure.go
```diff
@@ -0,0 +1,148 @@
...
+		State: StateSuccess,
+	}
```
csuzhangxc, line 39: init with `StateSuccess`, but `markCheckError` will not set it to `StateFailure`.
so, if `dbutil.GetTable` or `dbutil.GetCreateTableSQL` returns error (`c.errors == 0`), `r.State` still be `StateSuccess`.
change init value, or do `c.errors++`

pkg/check/table_structure.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/juju/errors"
+	"github.com/pingcap/tidb-tools/pkg/dbutil"
+	"github.com/pingcap/tidb/ast"
+	"github.com/pingcap/tidb/parser"
+	"github.com/pingcap/tidb/util/charset"
+)
+
+// TablesChecker checks table structures
+type TablesChecker struct {
```
csuzhangxc: any detail or ref?

pkg/check/table_structure.go
```diff
@@ -0,0 +1,148 @@
...
+	st, ok := stmt.(*ast.CreateTableStmt)
+	if !ok {
+		return errors.Errorf("Expect CreateTableStmt but got %T", stmt)
+	}
+	var err error
+
+	// check columns
+	for _, def := range st.Cols {
+		err = c.checkColumnDef(def)
+		if err != nil {
```
csuzhangxc, line 103: `c.checkColumnDef` has no effect? why add it

pkg/check/table_structure.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	// check options
+	for _, opt := range st.Options {
+		err = c.checkTableOption(opt)
+		if err != nil {
+			return errors.Trace(err)
+		}
+	}
+	return errors.Trace(err)
```
csuzhangxc: `err != nil` cases all returned before, so we need `return errors.Trace(err)` here?

pkg/check/table_structure.go
IANTHEREAL: leave a blank and fill it out later. If we check compatibility, we need to check definitions of column. constraints and table options in generally

pkg/check/table_structure.go
IANTHEREAL: not matter, leave it flexible.  Serious and lively

pkg/check/table_structure.go
IANTHEREAL: maybe habits of some people, errors.Trace(nil) = nil

dantin:
LGTM

csuzhangxc (approved):
LGTM

PR #44, opened 2018-07-17
checksum will be NULL if don't have any data

WangXiangUSTC:
don't need return error if checksum is NULL


WangXiangUSTC:
@GregoryIan @july2993 PTAL

IANTHEREAL:
LGTM

july2993 (approved):
LGTM

PR #50, opened 2018-08-14
basic implementation of sharding table checksum

holys:
1. calculate table checksum with crc64 independently (we will do comparison using another program since we have more than one sharding tables in different mysql instance.)
2. for a table, calculate md5 of each row, and then calculate the crc64 checksum of each md5 value with XOR.
3. for a table, we may split it into many chunks, so we do XOR with each chunk crc64 checksum
4. finally,  we do XOR with crc64 checksum of each sharding table.
5. the final checksum should equal to the checksum of merged table in tidb.


iamxy:
@WangXiangUSTC PTAL

IANTHEREAL:
@WangXiangUSTC  u issue a pr #60 , can we close this pr?

WangXiangUSTC:
ok, I will do the same thing in sync-diff-inspector, this pr can be close @holys @GregoryIan

PR #53, opened 2018-08-21
diff: use different checksum sql for tidb and mysql

WangXiangUSTC:
1. will judge the database's type by `select tidb_version`
2. use different checksum sql for tidb and mysql, and will get same checksum if data in tidb is equal to mysql's. (in the older tidb version, tidb will get different result with mysql, can see this issue https://github.com/pingcap/tidb/issues/7446)

### DIFF
--- a/pkg/dbutil/common.go
+++ b/pkg/dbutil/common.go
@@ -206,6 +206,8 @@ func GetCRC32Checksum(ctx context.Context, db *sql.DB, schemaName string, tbInfo
 		+------------+
 		| 1171947116 |
 		+------------+
+
+		Notice: in the older tidb version, tidb will get different checksum with mysql, can see this issue pingcap/tidb#7446
 	*/
 	isTiDB, err := IsTiDB(ctx, db)
 	if err != nil {
--- a/pkg/dbutil/common.go
+++ b/pkg/dbutil/common.go
@@ -33,6 +33,11 @@ const (
 	ImplicitColID = -1
 )
 
+var (
+	// ErrVersionNotFound means can't get the database's version
+	ErrVersionNotFound = errors.New("can't get the database's version")
+)
+
 // DBConfig is database configuration.
 type DBConfig struct {
 	Host string `toml:"host" json:"host"`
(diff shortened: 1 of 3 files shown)


WangXiangUSTC:
@GregoryIan @holys PTAL

sync_diff_inspector/main.go
```diff
@@ -27,6 +27,9 @@ import (
 	"github.com/pingcap/tidb-tools/pkg/utils"
 )
 
+var maxOpenConns = 100
```
IANTHEREAL: any error u meet, or set it to concurrency number?

pkg/dbutil/common.go
```diff
@@ -255,3 +280,56 @@ func SetSnapshot(ctx context.Context, db *sql.DB, snapshot string) error {
...
+	}
+
+	return "", errors.New("can't get the tidb's version")
+}
+
+// IsTiDB returns true if this database is tidb
+func IsTiDB(ctx context.Context, db *sql.DB) (bool, error) {
+	_, err := GetTiDBVersion(ctx, db)
+	if err != nil {
+		if strings.Contains(err.Error(), "tidb_version does not exist") {
```
IANTHEREAL: does mysql have error code that said `xxx  does not exist`?

sync_diff_inspector/main.go
WangXiangUSTC: meet `Cannot assign requested address` error

IANTHEREAL:
LGTM

pkg/dbutil/common.go
WangXiangUSTC: I don't find the error code

holys: mysql> select tidb_version1();
ERROR 1305 (42000): FUNCTION tidb_version1 does not exist

pkg/dbutil/common.go
```diff
@@ -255,3 +280,56 @@ func SetSnapshot(ctx context.Context, db *sql.DB, snapshot string) error {
...
+		}
+		break
+	}
+
+	if version.Valid {
+		// TODO: need analyse the tidb version string
+		return version.String, nil
+	}
+
+	return "", errors.New("can't get the tidb's version")
```
holys: Can we define an error for it?

pkg/dbutil/common.go
```diff
@@ -255,3 +280,56 @@ func SetSnapshot(ctx context.Context, db *sql.DB, snapshot string) error {
...
+}
+
+// IsTiDB returns true if this database is tidb
+func IsTiDB(ctx context.Context, db *sql.DB) (bool, error) {
```
holys: I think `select version() ` is better than `select tidb_version()`.

```
mysql> select version();
+--------------------------------------+
| version()                            |
+--------------------------------------+
| 5.7.10-TiDB-v2.1.0-beta-173-g7e48ab1 |
+--------------------------------------+
1 row in set (0.01 sec)
```
And then we can use `strings.Contains("5.7.10-TiDB-v2.1.0-beta-173-g7e48ab1", "TiDB")` to tell whether is tidb or not.

Ref: https://github.com/pingcap/tidb/commit/63ea1a8e8579ab83b4390a532da9161bdb982c9c

It's more reliable

WangXiangUSTC: done

pkg/dbutil/common.go
WangXiangUSTC: done

WangXiangUSTC:
@holys  PTAL again

holys:
LGTM

PR #62, opened 2018-09-06
pump client: remove strategy config

WangXiangUSTC:
remove strategy config, and this value will get from etcd. now use range as default


july2993 (approved):
LGTM

IANTHEREAL:
LGTM

PR #63, opened 2018-09-06
pump_client: remove log, print log to ioutil.Discard

WangXiangUSTC:
tidb team don't want print log from pump client, so modify these code.
similar with sarama, if want to out put the log,  do like 
https://github.com/pingcap/tidb-binlog/blob/57b629f9b1071ffef50e25bcca5a1fa2e534e125/drainer/util.go#L45

### DIFF
--- a/tidb-binlog/node/registry.go
+++ b/tidb-binlog/node/registry.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/coreos/etcd/clientv3"
 	"github.com/juju/errors"
-	"github.com/ngaut/log"
 	"github.com/pingcap/tidb-tools/pkg/etcd"
 	"golang.org/x/net/context"
 )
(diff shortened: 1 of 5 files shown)


IANTHEREAL:
@coocood PTAL

july2993:
currently, hard to tell if some code you use indirectly like in github.com/pingcap/tidb-tools/pkg/* will use log.Debug** or some day will, and the default level of `github.com/ngaut/log` is Debug

WangXiangUSTC:
what is your suggestion? @july2993

coocood:
Most of the log library have those methods:
```
Debugf
Infof
Warnf
Errorf
```
So if we define a logger interface that have those methods, whatever log lib TiDB use will be able to adapted to.

holys:
You may like this: https://github.com/siddontang/loggers   @WangXiangUSTC

coocood:
@holys 
Looks nice.
`Advanced` interface is what we need.

PR #71, opened 2018-09-13
pump client: use only one pump send commit binlog

WangXiangUSTC:
1. will not choose another pump for commit binlog if write failed
2. will retry 10 minute for commit binlog
3. commit binlog will choose the same pump with pre binlog, even though  the pump is not available.

### DIFF
--- a/tidb-binlog/pump_client/selector.go
+++ b/tidb-binlog/pump_client/selector.go
@@ -85,6 +85,12 @@ func (h *HashSelector) Select(binlog *pb.Binlog) *PumpStatus {
 	h.Lock()
 	defer h.Unlock()
 
+	if pump, ok := h.TsMap[binlog.StartTs]; ok {
+		// binlog is commit binlog or rollback binlog, choose the same pump by start ts map.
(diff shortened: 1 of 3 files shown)


WangXiangUSTC:
@july2993 @GregoryIan PTAL

tidb-binlog/pump_client/selector.go
```diff
@@ -175,30 +173,32 @@ func (r *RangeSelector) Select(binlog *pb.Binlog) *PumpStatus {
 		r.Unlock()
 	}()
 
+	if pump, ok := r.TsMap[binlog.StartTs]; ok {
```
july2993, line 176: does't C binlog never call Next ?

july2993:
LGTM

tidb-binlog/pump_client/selector.go
WangXiangUSTC: yes，next is only for pre binlog

IANTHEREAL:
LGTM

PR #74, opened 2018-09-27
diff: make diff a package && can ignore some columns

WangXiangUSTC:
1. make diff a package
2. can ignore some columns, will not check these columns' data
3. add remove columns config, will not check these columns' data, and will not use these columns as split field or order by key.
4. can ignore check table's struct or table's data


sync_diff_inspector/util.go
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+)
+
+func schemaStr(instanceID, schema string) string {
```
IANTHEREAL: is `schemaName` better?

pkg/diff/util.go
```diff
@@ -0,0 +1,129 @@
...
+			return false, nil
+		}
+	}
+
+	// check index
+	if len(tableInfo1.Indices) != len(tableInfo2.Indices) {
+		return false, nil
+	}
+
+	for i, index := range tableInfo1.Indices {
```
IANTHEREAL: is this check too strict？ do we need compatible with indexes in different orders？

pkg/diff/util.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	for i := 0; i < len(str1); i++ {
+		if str1[i] != str2[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func equalTableInfo(tableInfo1, tableInfo2 *model.TableInfo) (bool, error) {
```
IANTHEREAL: it's better to put it into dbutil pkg, check also does same thing.

pkg/diff/util.go
```diff
@@ -0,0 +1,129 @@
...
+			if col.Name.O != index2.Columns[j].Name.O {
+				return false, nil
+			}
+		}
+	}
+
+	return true, nil
+}
+
+func ignoreColumns(tableInfo *model.TableInfo, columns []string) *model.TableInfo {
```
IANTHEREAL: I don't think we need to change `tableinfo`,  e.g. a table with a unique key (id, name,age), if we remove id, name and age maybe not unique, how to sort them?

pkg/diff/util.go
```diff
@@ -0,0 +1,129 @@
...
+	if len(columns) == 0 {
+		return tableInfo
+	}
+	ignoreColMap := sliceToMap(columns)
+	for i, index := range tableInfo.Indices {
+		for j, col := range index.Columns {
+			if _, ok := ignoreColMap[col.Name.O]; ok {
+				index.Columns = append(index.Columns[:j], index.Columns[j+1:]...)
+				if len(index.Columns) == 0 {
+					tableInfo.Indices = append(tableInfo.Indices[:i], tableInfo.Indices[i+1:]...)
```
IANTHEREAL, line 51: put it out of for loop at L81

sync_diff_inspector/util.go
```diff
@@ -0,0 +1,30 @@
...
+	return fmt.Sprintf("%s|%s", instanceID, schema)
+}
+
+func sliceToMap(slice []string) map[string]interface{} {
```
IANTHEREAL: `pkg/diff/util` also has same function

pkg/diff/chunk.go
```diff
@@ -62,14 +62,14 @@ func newChunkRange(begin, end interface{}, containBegin, containEnd, noBegin, no
 		noEnd:        noEnd,
 	}
 }
-func getChunksForTable(db DBConfig, table *TableConfig, column *model.ColumnInfo, chunkSize, sample int) ([]chunkRange, error) {
+func getChunksForTable(table *TableInstance, column *model.ColumnInfo, chunkSize, sample int, limits string) ([]chunkRange, error) {
```
IANTHEREAL: unify name, s/limits/range

pkg/diff/diff.go
```diff
@@ -0,0 +1,532 @@
...
+	return structEqual, dataEqual, nil
+}
+
+// CheckTableStruct checks table's struct
+func (t *TableDiff) CheckTableStruct(ctx context.Context) (bool, error) {
+	tableInfo, err := dbutil.GetTableInfoWithRowID(ctx, t.TargetTable.Conn, t.TargetTable.Schema, t.TargetTable.Table, t.UseRowID)
+	if err != nil {
+		return false, errors.Trace(err)
+	}
+	log.Infof("")
```
IANTHEREAL: any useful?

IANTHEREAL:
@zhouqiang-cl @csuzhangxc PTAL

pkg/diff/chunk.go
WangXiangUSTC: range is a keyword in golang, so I use limits

pkg/diff/util.go
WangXiangUSTC: I add a remove-columns config, if user set ignore-columns, will not check these columns' data, if set remove-columns, will remove these columns from table info.

pkg/diff/util.go
WangXiangUSTC: done

pkg/diff/util.go
WangXiangUSTC: done

pkg/diff/util.go
WangXiangUSTC: only after remove columns, may index had 0 columns

pkg/dbutil/table.go
```diff
@@ -280,3 +280,48 @@ func setImplicitColumn(table *model.TableInfo) {
 	}
 	table.Indices = []*model.IndexInfo{newIndex}
 }
+
+// EqualTableInfo returns true if this two table info have same columns and indices
+func EqualTableInfo(tableInfo1, tableInfo2 *model.TableInfo) (bool, error) {
```
zhouqiang-cl: when return error

pkg/diff/util.go
```diff
@@ -0,0 +1,92 @@
...
+
+	for j := 0; j < num; j++ {
+		r := j + rand.Intn(total-j)
+		totalArray[j], totalArray[r] = totalArray[r], totalArray[j]
+	}
+
+	return totalArray[:num]
+}
+
+func needQuotes(ft types.FieldType) bool {
```
zhouqiang-cl, line 88: number & float can use quota too

sync_diff_inspector/util.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+)
+
+func schemaName(instanceID, schema string) string {
```
zhouqiang-cl, line 20: where use this function

zhouqiang-cl (commented):
rest LGTM

pkg/dbutil/table.go
WangXiangUSTC: I will remove error

pkg/diff/util.go
WangXiangUSTC: I will remove this function

sync_diff_inspector/util.go
WangXiangUSTC: it is used for generate a label for schema,  you can search the code

pkg/diff/chunk.go
csuzhangxc: so, is it possible to put `limits` into `table`

pkg/diff/util.go
WangXiangUSTC: but "2"> "11" will return true, I will keep this function

pkg/diff/chunk.go
WangXiangUSTC: table is a  instance for table, seems not suitable

pkg/diff/util.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"math/rand"
+
+	"github.com/ngaut/log"
+	"github.com/pingcap/tidb-tools/pkg/dbutil"
+	"github.com/pingcap/tidb/model"
+	"github.com/pingcap/tidb/types"
+)
+
+func equalStrings(str1, str2 []string) bool {
```
csuzhangxc, line 25: where use this function? and why not use `==` simply

WangXiangUSTC: slice can only be compared to nil. this function is used to check whether two schema have same tables, I will use this function later

csuzhangxc: wow, my mistake

pkg/diff/util.go
```diff
@@ -0,0 +1,92 @@
...
+
+func removeColumns(tableInfo *model.TableInfo, columns []string) *model.TableInfo {
+	if len(columns) == 0 {
+		return tableInfo
+	}
+	removeColMap := SliceToMap(columns)
+	for i, index := range tableInfo.Indices {
+		for j, col := range index.Columns {
+			if _, ok := removeColMap[col.Name.O]; ok {
+				index.Columns = append(index.Columns[:j], index.Columns[j+1:]...)
```
csuzhangxc, line 48: is is possible `slice bounds out of range` for `index.Columns[j+1:]` ?

csuzhangxc:
rest LGTM

zhouqiang-cl (approved):
LGTM

pkg/diff/util.go
WangXiangUSTC: yes, this is a bug, I fix it

pkg/diff/util.go
```diff
@@ -38,21 +38,28 @@ func removeColumns(tableInfo *model.TableInfo, columns []string) *model.TableInf
 	if len(columns) == 0 {
 		return tableInfo
 	}
+
 	removeColMap := SliceToMap(columns)
-	for i, index := range tableInfo.Indices {
-		for j, col := range index.Columns {
+	for i := 0; i < len(tableInfo.Indices); i++ {
```
IANTHEREAL, line 43: I prefer to construct a new columns. but we can keep it

IANTHEREAL:
LGTM

PR #84, opened 2018-10-19
diff: minor update about collation

WangXiangUSTC:


IANTHEREAL:
what error u met?

pkg/dbutil/common.go
```diff
@@ -142,7 +142,7 @@ func GetRowCount(ctx context.Context, db *sql.DB, schemaName string, tableName s
 func GetRandomValues(ctx context.Context, db *sql.DB, schemaName, table, column string, num int64, min, max interface{}, limitRange string, collation string) ([]interface{}, error) {
 	/*
 		example:
-		mysql> SELECT `id` FROM (SELECT `id` FROM `test`.`test` WHERE `id` COLLATE "latin1_bin" > 0 AND `id` COLLATE "latin1_bin" < 100 AND true ORDER BY RAND() LIMIT 3)rand_tmp ORDER BY `id`;
+		mysql> SELECT `id` FROM (SELECT `id` FROM `test`.`test` WHERE `id` COLLATE "latin1_bin" > 0 AND `id` COLLATE "latin1_bin" < 100 AND true ORDER BY RAND() LIMIT 3)rand_tmp ORDER BY `id` COLLATE "latin1_bin";
```
WangXiangUSTC, line 145: we use this function to get some values to split chunk, if without this config in sort, may split wrong chunk. 
For example, we set collate="latin1_bin", the old sql get three values:
v1, v2, v3
use these value split two chunks: [v1, v2), [v2, v3]
but under collate  "latin1_bin", v3 is less than v2, [v2, v3] don't contain any data.
so we should add collate config in order by.

pkg/diff/chunk.go
```diff
@@ -87,13 +87,14 @@ func getChunksForTable(table *TableInstance, column *model.ColumnInfo, chunkSize
 
 	field := column.Name.O
 
+	collationStr := ""
 	if collation != "" {
-		collation = fmt.Sprintf(" COLLATE \"%s\"", collation)
+		collationStr = fmt.Sprintf(" COLLATE \"%s\"", collation)
```
WangXiangUSTC, line 92: collation is reused in  L134, should not changed it's value, otherwise will generate a wrong sql

WangXiangUSTC (commented):
PTAL again @GregoryIan

IANTHEREAL:
LGTM

PR #116, opened 2018-11-12
update go-sql-driver's version

WangXiangUSTC:
I update tidb-tools' version in tidb, but ci failed, https://github.com/pingcap/tidb/pull/8265. it is because tidb and tools use different go-sql-driver, and tools appoint the version. need use same go-sql-driver version with tidb.

go mod edit -require=github.com/go-sql-driver/mysql@3955978caca48c1658a4bb7a9c6a0f084e326af3
make

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/eknkc/amber v0.0.0-20171010120322-cdade1c07385 // indirect
 	github.com/etcd-io/gofail v0.0.0-20180808172546-51ce9a71510a // indirect
 	github.com/ghodss/yaml v1.0.0 // indirect
-	github.com/go-sql-driver/mysql v1.4.0
+	github.com/go-sql-driver/mysql v0.0.0-20170715192408-3955978caca4
 	github.com/gogo/protobuf v1.1.1 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/eknkc/amber v0.0.0-20171010120322-cdade1c07385 // indirect
 	github.com/etcd-io/gofail v0.0.0-20180808172546-51ce9a71510a // indirect
 	github.com/ghodss/yaml v1.0.0 // indirect
-	github.com/go-sql-driver/mysql v1.4.0
+	github.com/go-sql-driver/mysql v0.0.0-20170715192408-3955978caca4
```
IANTHEREAL, line 23: newer version？ in fact we should test it

kennytm:
This... doesn't quite make sense. The commit go-sql-driver/mysql@3955978caca48c1658a4bb7a9c6a0f084e326af3 is dated 2017 Jul 16th, and the release [v1.4.0](https://github.com/go-sql-driver/mysql/commits/v1.4.0) is on 2018 Jun 3rd, there is no reason *we* want to downgrade the dependency from a newer released version to an older anonymous version.

WangXiangUSTC:
@kennytm  I will close this pr, and create a pr to rename go.mod to go.mod1

PR #122, opened 2018-11-15
diff: fix sourcetable in table struct diff

### LINKED ISSUE #121: 【Question】tidb-tools/pkg/diff/diff.go
* target table info from:
https://github.com/pingcap/tidb-tools/blob/d57120b76193f2b462df38d080a9b734cb6c2485/pkg/diff/diff.go#L124
*source table info from:
https://github.com/pingcap/tidb-tools/blob/d57120b76193f2b462df38d080a9b734cb6c2485/pkg/diff/diff.go#L131

`Source and target table generated by the same connection, schema ? `

amyangfei:
fix https://github.com/pingcap/tidb-tools/issues/121

Use right sourceTable to get tableInfo

### Check List 

Tests 

   * create table t1 in both source db and target db, with different ddl.
   * use sync_diff_inspector to check and get `table's struct not equal`

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -128,7 +128,7 @@ func (t *TableDiff) CheckTableStruct(ctx context.Context) (bool, error) {
 	t.TargetTable.info = removeColumns(tableInfo, t.RemoveColumns)
 
 	for _, sourceTable := range t.SourceTables {
-		tableInfo, err := dbutil.GetTableInfoWithRowID(ctx, t.TargetTable.Conn, t.TargetTable.Schema, t.TargetTable.Table, t.UseRowID)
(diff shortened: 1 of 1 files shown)


WangXiangUSTC:
LGTM

jianhaiqing:
Is it possible to add test for this kind of case ? It actually bothers me when it comes to test of connecting MySQL,  mock it ? how ?

WangXiangUSTC:
will add ci test for this project later @jianhaiqing

IANTHEREAL:
good catch! @amyangfei   
please add unit tests for diff tool ASAP @WangXiangUSTC

IANTHEREAL:
LGTM

PR #137, opened 2018-11-30
Add timezone checker between upstream/downstream db

amyangfei:
Add timezone checker between upstream/downstream db, used by https://internal.pingcap.net/jira/browse/TOOL-653

Add a checker and some timezone query function in dbuil

### Check List 

Tests 

 - Integration test
https://internal.pingcap.net/jira/browse/TOOL-653

Code changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 deps.sh
 go.sum
 go.mod
+vendor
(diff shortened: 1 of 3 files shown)


amyangfei:
TiDB doesn't provide a right `system_time_zone` value, so the check is not always accurate. Use other method to solve this issue. 

```
InferSystemTZ reads system timezone from `TZ`, the path of the soft link of `/etc/localtime`. If both of them are failed, system timezone will be set to `UTC`.
// It is exported because we need to use it during bootstap stage. And it should be only used at that stage.
```

PR #150, opened 2018-12-25
use old version pd, and remove ddl checker temporary

WangXiangUSTC:
use old version pd, for compatible with tidb 2.0

and ddl checker may relay on the newest version tidb, remove it temporary.

use old version pd, update pd/client to pd/pd-client
rm ddl checker


IANTHEREAL:
LGTM

july2993 (approved):

PR #160, opened 2019-01-07
diff: refine code && add unit test for sharding tables

WangXiangUSTC:
1. refine code, rows should be closed after select, or it may cause blocked in next select.
2. add unit test for sharding tables

1. scan rows after select data, and then close rows
2. diff more than one source table with one target table in unit test

### Check List 

Tests 

 - Unit test
 - Integration test


pkg/dbutil/index.go
```diff
@@ -42,24 +42,24 @@ func ShowIndex(ctx context.Context, db *sql.DB, schemaName string, table string)
 	defer rows.Close()
 
 	for rows.Next() {
-		fields, _, err1 := ScanRow(rows)
+		fields, err1 := ScanRow(rows)
 		if err1 != nil {
 			return nil, errors.Trace(err1)
 		}
-		seqInIndex, err1 := strconv.Atoi(string(fields["Seq_in_index"]))
+		seqInINdex, err1 := strconv.Atoi(string(fields["Seq_in_index"].Data))
```
kennytm:
```suggestion
		seqInIndex, err1 := strconv.Atoi(string(fields["Seq_in_index"].Data))
```

pkg/diff/chunk.go
```diff
@@ -201,10 +201,6 @@ func (s *randomSpliter) split(table *TableInstance, columns []*model.ColumnInfo,
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	if cnt == 0 {
-		log.Infof("no data found in %s.%s", table.Schema, table.Table)
-		return nil, nil
```
WangXiangUSTC, line 206: if count is 0, also generate a chunk for this table

WangXiangUSTC:
@GregoryIan @csuzhangxc  PTAL

pkg/diff/diff_test.go
```diff
@@ -246,16 +270,46 @@ func generateData(dbConn *sql.DB, dbCfg dbutil.DBConfig, sourceTable, targetTabl
...
 	if err != nil {
 		return err
 	}
 
+	conditions := make([]string, 0, 3)
+	if randomValueNum == 0 {
+		conditions = append(conditions, "true")
+	} else {
+		conditions = append(conditions, fmt.Sprintf("e < %s", values[0]))
+		for i := range values {
```
csuzhangxc: `for i := 0; i < len(values)-1; i++ {`, then no `break` needed?

pkg/diff/diff_test.go
```diff
@@ -246,16 +270,46 @@ func generateData(dbConn *sql.DB, dbCfg dbutil.DBConfig, sourceTable, targetTabl
 	importer.DoProcess(cfg)
 
 	// generate data for target table
```
csuzhangxc: update this comment.

WangXiangUSTC:
@csuzhangxc PTAL again

csuzhangxc:
LGTM

pkg/diff/chunk.go
IANTHEREAL: why？

pkg/diff/diff.go
```diff
@@ -298,13 +299,11 @@ func (t *TableDiff) checkChunkDataEqual(ctx context.Context, checkJobs []*CheckJ
 	return equal, nil
 }
 
-func (t *TableDiff) compareRows(sourceRows map[string]*sql.Rows, targetRows *sql.Rows, orderKeyCols []*model.ColumnInfo) (bool, error) {
+func (t *TableDiff) compareRows(sourceRows map[string][]map[string]*dbutil.ColumnData, targetRows []map[string]*dbutil.ColumnData, orderKeyCols []*model.ColumnInfo) (bool, error) {
```
IANTHEREAL, line 302: why change it？

pkg/diff/chunk.go
WangXiangUSTC: empty in target, but may be have data in source

pkg/diff/diff.go
WangXiangUSTC: we should read data from sql.Rows and then close the rows.

IANTHEREAL: if use `rows`, we can read it while we need, now we should read all once, right?

but you can keep it, while you need to refine mem usage, you should change it back

IANTHEREAL:
LGTM

pkg/diff/diff.go
WangXiangUSTC: ok

PR #165, opened 2019-01-10
pump client: add initial log function

WangXiangUSTC:
need print pump client's log in tidb, use tidb's log config.
https://internal.pingcap.net/jira/browse/TOOL-764

use tidb's log config function

### Check List 

Tests 

 - Manual test

### DIFF
--- a/tidb-binlog/pump_client/client.go
+++ b/tidb-binlog/pump_client/client.go
@@ -28,6 +28,7 @@ import (
 	"github.com/pingcap/tidb-tools/pkg/etcd"
 	"github.com/pingcap/tidb-tools/pkg/utils"
 	"github.com/pingcap/tidb-tools/tidb-binlog/node"
+	"github.com/pingcap/tidb/util/logutil"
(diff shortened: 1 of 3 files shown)


WangXiangUSTC:
@GregoryIan @july2993 PTAL

tidb-binlog/pump_client/client.go
```diff
@@ -49,7 +50,7 @@ const (
 )
 
 var (
-	// Logger is ...
+	// Logger is ..., obsolete now.
 	Logger = log.New()
```
july2993, line 54: just delete it if no use

july2993 (commented):
rest LGTM

tidb-binlog/pump_client/client.go
WangXiangUSTC: need update tidb's code, and then I will clean all obsolete fields

IANTHEREAL:
LGTM

july2993 (approved):
LGTM

PR #168, opened 2019-01-10
adjust a few error messages

lewgun:
adjust a few error messages

only adjust a few error messages, don't change any logic

### Check List 

Tests 

 - Unit test

Code changes

- N/A

Side effects

- N/A

Related changes
- N/A

### DIFF
--- a/pkg/column-mapping/column.go
+++ b/pkg/column-mapping/column.go
@@ -465,7 +465,7 @@ func partitionID(info *mappingInfo, vals []interface{}) ([]interface{}, error) {
 	}
 
 	if originID >= maxOriginID || originID < 0 {
-		return nil, errors.NotValidf("id must less than %d, bigger or equal than 0, but get %d", maxOriginID, originID)
+		return nil, errors.NotValidf("id must less than %d, greater than or equal to 0, but got %d", maxOriginID, originID)
 	}
 
 	originID = int64(info.instanceID | info.schemaID | info.tableID | originID)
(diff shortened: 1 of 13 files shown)


WangXiangUSTC:
@lewgun  thanks, LGTM

pkg/column-mapping/column.go
```diff
@@ -507,13 +507,13 @@ func computePartitionID(schema, table string, rule *Rule) (instanceID int64, sch
...
 	if len(prefix) >= len(name) || prefix != name[:len(prefix)] {
-		return 0, errors.NotValidf("%s is not prefix of %s", prefix, name)
+		return 0, errors.NotValidf("%s is not the prefix of %s", prefix, name)
 	}
 
 	idStr := name[len(prefix):]
 	id, err := strconv.ParseUint(idStr, 10, bitSize)
 	if err != nil {
-		return 0, errors.NotValidf("suffix of %s is not int64", idStr)
+		return 0, errors.NotValidf("suffix of %s's data type is not int64", idStr)
```
csuzhangxc: we should focus on `idStr`, can we change the statement not followed by `'s`?

lewgun: like this :"suffix of the idStr, can't be converted to int64" ?

csuzhangxc: `suffix of the idStr can't be converted to int64`?

lewgun: LGTM

lewgun:
PTAL @csuzhangxc

csuzhangxc (approved):
LGTM

PR #174, opened 2019-01-15
pkg: added the "filter" package

kennytm:

Added the "filter" which performs schema/table filtering for DM and Lightning. The package implements the MySQL replication rules. Since it is shared among multiple projects, it is better to move the package into this common repository.

Moved the existing package from DM to here.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/filter/README.md
+++ b/pkg/filter/README.md
@@ -1,0 +1,17 @@
+## Filter  
+
+filter provides a library to filter replicate on schema/table by given rules
+
+### Rules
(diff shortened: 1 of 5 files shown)


kennytm:
PTAL @GregoryIan @csuzhangxc

amyangfei (commented):
after this pr is merged, remember to provide another pr in DM repo
LGTM

lonng (approved):
LGTM

PR #178, opened 2019-01-16
diff: minor update on table route

WangXiangUSTC:
for example, we set
[[table-rules]]
schema-pattern = "test1"
target-schema = "test2"

we only want compare schema test1 on source with schema test2 on target, but now will also compare schema test2 on source, because route("test2") return "test2" although test2 mismatched the rules.

return matched when execute route

### Check List 

Tests 

 - Manual test (use config in example above)

### DIFF
--- a/pkg/table-router/router.go
+++ b/pkg/table-router/router.go
@@ -123,7 +123,7 @@ func (r *Table) RemoveRule(rule *TableRule) error {
 
 // Route routes schema/table to target schema/table
 // don't support to route schema/table to multiple schema/table
-func (r *Table) Route(schema, table string) (string, string, error) {
+func (r *Table) Route(schema, table string) (targetSchema string, targetTable string, matched bool, err error) {
 	schemaL, tableL := schema, table
 	if !r.caseSensitive {
 		schemaL, tableL = strings.ToLower(schema), strings.ToLower(table)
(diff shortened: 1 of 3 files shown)


IANTHEREAL:
I don't understand your description. do you want to compare source `test2` and target `test2`, or not?

WangXiangUSTC:
don't need compare source test2 with target test2 @GregoryIan

csuzhangxc:
LGTM. you also implement this in #130, right?

WangXiangUSTC:
yes, I revert it in #130 now @csuzhangxc

pkg/table-router/router.go
```diff
@@ -123,7 +123,7 @@ func (r *Table) RemoveRule(rule *TableRule) error {
 
 // Route routes schema/table to target schema/table
 // don't support to route schema/table to multiple schema/table
-func (r *Table) Route(schema, table string) (string, string, error) {
+func (r *Table) Route(schema, table string) (targetSchema string, targetTable string, matched bool, err error) {
```
IANTHEREAL, line 126: it would affect `syncer`/`loader`/`dm`, I wouldn't like to change it like this, if you need to filter `test2`, just give a bw list

WangXiangUSTC: dm can just ignore the return value `matched`

IANTHEREAL: it's very strange, some one would think you should compare `test2` and `test2`

PR #179, opened 2019-01-16
simplify the if/else logic

lewgun:
simplify the if/else logic

refactor to simplify the if/else logic

### Check List 

Tests 

 - Unit test

Code changes

- N/A

Side effects

- N/A

Related changes

 - N/A

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -506,29 +506,34 @@ func compareData(map1, map2 map[string][]byte, null1, null2 map[string]bool, ord
 			return false, 0, errors.Errorf("don't have key %s", col.Name.O)
 		}
 		if needQuotes(col.FieldType) {
-			if string(data1) > string(data2) {
-				cmp = 1
-				break
-			} else if string(data1) < string(data2) {
-				cmp = -1
-				break
-			} else {
+
+			strData1 := string(data1)
+			strData2 := string(data2)
+
+			if strData1 == strData2 {
 				continue
 			}
+
+			cmp = -1
+			if strData1 > strData2 {
+				cmp = 1
+			}
+			break
+
 		} else {
 			num1, err1 := strconv.ParseFloat(string(data1), 64)
 			num2, err2 := strconv.ParseFloat(string(data2), 64)
 			if err1 != nil || err2 != nil {
 				return false, 0, errors.Errorf("convert %s, %s to float failed, err1: %v, err2: %v", string(data1), string(data2), err1, err2)
 			}
+
+			if num1 == num2 {
+				continue
+			}
+
+			cmp = -1
 			if num1 > num2 {
 				cmp = 1
-				break
-			} else if num1 < num2 {
-				cmp = -1
-				break
-			} else {
-				continue
 			}
 		}
 	}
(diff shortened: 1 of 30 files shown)


pkg/diff/merge.go
```diff
@@ -55,19 +54,25 @@ func (r RowDatas) Less(i, j int) bool {
 				}
 				continue
 			}
+			if strData1 > strData2 {
+				return false
+			}
+			return true
 		} else {
```
amyangfei: recommend usage from golint: `if block ends with a return statement, so drop this else and outdent its block`

amyangfei:
Good job. Rest LGTM

lewgun:
@amyangfei  PTAL

amyangfei:
LGTM

IANTHEREAL:
Thanks @lewgun, @WangXiangUSTC PTAL

pkg/diff/diff.go
```diff
@@ -506,30 +506,36 @@ func compareData(map1, map2 map[string][]byte, null1, null2 map[string]bool, ord
...
-				break
-			} else if string(data1) < string(data2) {
-				cmp = -1
-				break
-			} else {
+
+			strData1 := string(data1)
+			strData2 := string(data2)
+
+			if strData1 == strData2 {
```
lonng: I think it is better to compare string length first: `if len(strData1) == len(strData2) && ...`

lewgun: done

pkg/diff/merge.go
```diff
@@ -41,11 +41,10 @@ func (r RowDatas) Less(i, j int) bool {
...
 		if needQuotes(col.FieldType) {
-			if string(data1) > string(data2) {
-				return false
-			} else if string(data1) < string(data2) {
-				return true
-			} else {
+			strData1 := string(data1)
+			strData2 := string(data2)
+
+			if strData1 == strData2 {
```
lonng, line 47: `len(strData1) == len(strData2) && ..`

lonng (approved):
Rest LGTM

pkg/diff/merge.go
kennytm: @lonng  I disagree with the `len(a) == len(b) && a == b` check because there is no algorithmic advantage at all. Firstly, `a == b` already does the length check (see golang/go#1161).

Now, if `len(a) == len(b)`, then `a == b` will do the O(n) comparison anyway. If `len(a) != len(b)` then it will still perform the check `strData1 > strData2` which is O(n) again.

According to the comment [in `strings.Compare`](https://golang.org/src/strings/compare.go), the pattern 

```go
if a == b {
    return foo;
}
if a < b {
    return bar;
}
return bar;
```

will always be optimized into a [single](https://github.com/golang/go/blob/master/src/internal/bytealg/compare_generic.go#L39-L60) O(n) string comparison 

> NOTE(rsc): This function does NOT call the runtime cmpstring function,
> because we do not want to provide any performance justification for
> using strings.Compare. Basically no one should use strings.Compare.
> As the comment above says, it is here only for symmetry with package bytes.
> **If performance is important, the compiler should be changed to recognize
> the pattern so that all code doing three-way comparisons**, not just code
> using strings.Compare, can benefit.

IMO all those `len(a) == len(b) &&` check should be removed unless there's a performance test showing it is indeed faster.

lonng: OK

PR #182, opened 2019-01-17
DM: relation between `source_id` and `instance-id`

csuzhangxc:

- declare how to set `source_id` to keep using checkpoints

- add more descriptions

### Check List 

Tests 

 - No code


amyangfei (commented):
LGTM

PR #191, opened 2019-01-29
simplify the if/else logic

lewgun:
**What problem does this PR solve?**

simplify the if/else logic

**What is changed and how it works?**

refactor to simplify the if/else logic

**Check List**

Tests

- Unit test

Code changes

- N/A

Side effects

- N/A

Related changes

- N/A

### DIFF
--- a/pkg/diff/chunk.go
+++ b/pkg/diff/chunk.go
@@ -86,7 +86,7 @@ func (c *chunkRange) toString(mode string, collation string) (string, []string) 
 
 	/* for example:
 	there is a bucket in TiDB, and the lowerbound and upperbound are (v1, v3), (v2, v4), and the columns are `a` and `b`,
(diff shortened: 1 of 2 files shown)


pkg/diff/chunk.go
```diff
@@ -243,25 +243,26 @@ func (s *randomSpliter) splitRange(db *sql.DB, chunk *chunkRange, count int, sch
 		symbolMax = chunk.bounds[colNum-1].upperSymbol
 	} else {
 		// choose the next column to split data
```
WangXiangUSTC: please move this comment to L250

WangXiangUSTC (commented):
thanks! rest LGTM

pkg/diff/chunk.go
lewgun: done PTAL @WangXiangUSTC

WangXiangUSTC:
LGTM

Issue #187, opened 2019-01-24
panic: runtime error: invalid memory address or nil pointer dereference


bqwzhl20:
2019/01/24 11:19:50 diff.go:258: [error] table: test_user, range: (`uid` >= ? AND `uid` < ? AND ctime >= '2017-05-09 10:39:37' AND ctime < '2019-01-23 17:40:00'), args: [151262720433299085 151753833161977213], checksum is not equal, one is 1422990440, another is 2838034901
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x68 pc=0xc5fcfe]

goroutine 29 [running]:
github.com/pingcap/tidb-tools/pkg/dbutil.SelectUniqueOrderKey(0x0, 0xed0060, 0x11, 0xc000147b18, 0x4c6816, 0xc0001ee840, 0x0)
        /home/jenkins/workspace/build_tidb_tools_master/go/src/github.com/pingcap/tidb-tools/pkg/dbutil/index.go:165 +0xae
github.com/pingcap/tidb-tools/pkg/diff.getChunkRows(0xfcda60, 0xc0000ac010, 0xc0001ee9c0, 0xc0002dbb00, 0xf, 0xc0002dbb40, 0x10, 0x0, 0xc00043cd20, 0x67, ...)
        /home/jenkins/workspace/build_tidb_tools_master/go/src/github.com/pingcap/tidb-tools/pkg/diff/diff.go:541 +0x43
github.com/pingcap/tidb-tools/pkg/diff.(*TableDiff).checkChunkDataEqual(0xc0000da480, 0xfcda60, 0xc0000ac010, 0xc0004f8000, 0x419, 0x1063, 0x0, 0x0, 0x0)
        /home/jenkins/workspace/build_tidb_tools_master/go/src/github.com/pingcap/tidb-tools/pkg/diff/diff.go:264 +0x19f
github.com/pingcap/tidb-tools/pkg/diff.(*TableDiff).EqualTableData.func1(0xc0000da480, 0xfcda60, 0xc0000ac010, 0xc0004cf3b0, 0xc0004f8000, 0x419, 0x1063)
        /home/jenkins/workspace/build_tidb_tools_master/go/src/github.com/pingcap/tidb-tools/pkg/diff/diff.go:192 +0x73
created by github.com/pingcap/tidb-tools/pkg/diff.(*TableDiff).EqualTableData
[...]

PR #195, opened 2019-02-11
diff: get table info before check

WangXiangUSTC:
issue: https://github.com/pingcap/tidb-tools/issues/187
when `ignore-struct-check = true` or some table's struct is not equal, some table's info may be nil.

get all table's info when adjust config in source and target.

### Check List 

Tests 

 - Integration test
 - Manual test (set ignore-struct-check = true)

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -93,10 +93,7 @@ type TableDiff struct {
 
 // Equal tests whether two database have same data and schema.
 func (t *TableDiff) Equal(ctx context.Context, writeFixSQL func(string) error) (bool, bool, error) {
-	err := t.adjustConfig(ctx)
-	if err != nil {
-		return false, false, errors.Trace(err)
-	}
+	t.adjustConfig()
 
 	t.sqlCh = make(chan string)
 	t.wg.Add(1)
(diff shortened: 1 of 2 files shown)


pkg/diff/diff.go
```diff
@@ -163,6 +154,26 @@ func (t *TableDiff) adjustConfig() {
 	if t.CheckThreadCount <= 0 {
 		t.CheckThreadCount = 4
 	}
+
+	return t.getTableInfo(ctx)
```
csuzhangxc: I think calling this function in `Equal` following `adjustConfig` is better, so `adjustConfig` only adjusts the config.

csuzhangxc (commented):
rest LGTM

csuzhangxc:
LGTM

amyangfei:
LGTM

amyangfei:
Is CI for this repo missing now? PTAL @zhouqiang-cl

IANTHEREAL:
any CI testings？ @WangXiangUSTC

WangXiangUSTC:
I have already request @cwen0 for help @GregoryIan

PR #198, opened 2019-02-13
driver/reader: return the next offset instead of sarama.OffsetNewest

july2993:
relate issue: https://internal.pingcap.net/jira/browse/TOOL-932
Suppose the ts in queue is
1 2 3
we use ts = 3 to seek the offset, it return sarama.OffsetNewest
at the same time some data pushed to queue it become
1 2 3 4
we use sarama.OffsetNewest to consume, so we lost the msg ts = 4

return the next offset instead of sarama.OffsetNewest

### Check List 

Tests 

Code changes

Side effects

Related changes


WangXiangUSTC:
LGTM

IANTHEREAL:
LGTM

PR #209, opened 2019-03-01
diff: add unit test for chunk split

WangXiangUSTC:
add unit test for chunk split

1. move code about split to spliter.go
2. use mocksql to create fake data, and then check the chunk is expected or not.
3. fix some corner case

### Check List 

Tests 

 - Unit test


pkg/diff/chunk.go
```diff
@@ -332,36 +340,37 @@ func (s *randomSpliter) splitRange(db *sql.DB, chunk *chunkRange, count int, sch
...
-				if useNewColumn {
-					lower = ""
-					lowerSymbol = ""
-				} else {
-					lower = splitValues[i]
-					lowerSymbol = symbolMin
-				}
+		}
+
+		if i == 0 && valueCounts[i] == 1 {
```
IANTHEREAL: it's very unfriendly to reviewer, please refine these code, make the logic more clear

pkg/diff/spliter_test.go
```diff
@@ -0,0 +1,271 @@
...
+
+import (
+	"fmt"
+
+	sqlmock "github.com/DATA-DOG/go-sqlmock"
+	. "github.com/pingcap/check"
+	"github.com/pingcap/tidb-tools/pkg/dbutil"
+)
+
+var _ = Suite(&testSpliterSuite{})
```
IANTHEREAL, line 24: Good Job!

WangXiangUSTC:
@GregoryIan PTAL again

Makefile
```diff
@@ -55,9 +57,7 @@ fmt:
 check:
 	#go get github.com/golang/lint/golint
 	@echo "vet"
-	@ go tool vet $(FILES) 2>&1 | awk '{print} END{if(NR>0) {exit 1}}'
-	@echo "vet --shadow"
-	@ go tool vet --shadow $(FILES) 2>&1 | awk '{print} END{if(NR>0) {exit 1}}'
+	$(GO) vet -all $(PACKAGES) 2>&1 | tee /dev/stderr | $(FAIL_ON_STDOUT)
```
WangXiangUSTC, line 60: update go version to 1.12, execute `go tool vet` will return error `Invoking "go tool vet" directly is unsupported; use "go vet"`

csuzhangxc (commented):
LGTM

IANTHEREAL:
LGTM

PR #212, opened 2019-03-05
refine name of shard table checker

IANTHEREAL:
refine `Name()` of shard table checker


WangXiangUSTC:
LGTM

PR #215, opened 2019-03-14
diff: add config to only use checksum, skip select data

WangXiangUSTC:
in some case, user only want to check data is equal or not, and don't want to know which row is not equal. 

add a config `only-use-checksum`, if set to true, will skip select data when checksum is not equal.

### Check List 

Tests 

 - Manual test 

Code changes

Related changes


sync_diff_inspector/config.go
```diff
@@ -312,5 +315,16 @@ func (c *Config) checkConfig() bool {
 		}
 	}
 
+	if c.OnlyUseChecksum {
+		if !c.UseChecksum {
+			log.Error("need set use-checksum = true")
```
csuzhangxc, line 321: whether needing to return an error?

another question, should we need to move some config check into `diff` pkg?

sync_diff_inspector/config.toml
```diff
@@ -18,15 +18,20 @@ sample-percent = 100
...
+
 # ignore check table's data
 ignore-data-check = false
 
 # ignore check table's struct
 ignore-struct-check = false
 
-# the name of the file which saves sqls used to fix different data
+# the name of the file which saves sqls used to fix different data.
+# if fix-sql-file is "", will only check data by checksum.
```
csuzhangxc: where do we implement this?

sync_diff_inspector/config.go
WangXiangUSTC: 1. add return false
2. can also add some config check in diff pkg, will do this later

sync_diff_inspector/config.toml
WangXiangUSTC: forget update comment, done

csuzhangxc (commented):
LGTM

pkg/diff/diff_test.go
```diff
@@ -214,7 +214,7 @@ func testDataEqual(dbConn *sql.DB, sourceTables []string, targetTable string, ha
 		_, err = dbConn.Exec(sql)
 		c.Assert(err, IsNil)
 	}
-	structEqual, dataEqual, err = tableDiff.Equal(context.Background(), writeSqls)
+	structEqual, dataEqual, err = tableDiff.Equal(context.Background(), nil)
```
amyangfei: why change `writeSqls` to nil here.
if checked data is equal, there is no side effect. But if checked data is not equal (of course the inequality is caused by bug, but during development it may happen), the test will panic.

sync_diff_inspector/config.go
```diff
@@ -312,5 +315,17 @@ func (c *Config) checkConfig() bool {
...
+			return false
+		}
+	} else {
+		if len(c.FixSQLFile) == 0 {
```
IANTHEREAL, line 325: does it have a default value?

sync_diff_inspector/config.go
IANTHEREAL: do it

pkg/diff/diff_test.go
WangXiangUSTC: done

sync_diff_inspector/config.go
WangXiangUSTC: set default value here

IANTHEREAL:
LGTM

PR #217, opened 2019-03-17
update log to pingcap/log

WangXiangUSTC:
update log to pingcap/log

update log pkg

### Check List 

Tests 

 - Unit test

Related changes


pkg/dbutil/common.go
```diff
@@ -243,7 +244,7 @@ func GetMinMaxValue(ctx context.Context, db *sql.DB, schema, table, column strin
 
 	query := fmt.Sprintf("SELECT /*!40001 SQL_NO_CACHE */ MIN(`%s`%s) as MIN, MAX(`%s`%s) as MAX FROM `%s`.`%s` WHERE %s",
 		column, collation, column, collation, schema, table, limitRange)
-	log.Debugf("GetMinMaxValue query: %v, args: %v", query, limitArgs)
+	log.Debug("GetMinMaxValue", zap.String("sql", query), zap.String("args", fmt.Sprintf("%v", limitArgs)))
```
kennytm: Perhaps use `zap.Reflect("args", limitArgs)` here?

pkg/diff/diff.go
```diff
@@ -217,7 +218,7 @@ func (t *TableDiff) EqualTableData(ctx context.Context) (equal bool, err error)
 		go func(checkJobs []*CheckJob) {
 			eq, err := t.checkChunkDataEqual(ctx, checkJobs)
 			if err != nil {
-				log.Errorf("check chunk data equal failed, error %v", errors.ErrorStack(err))
+				log.Error("check chunk data equal failed", zap.String("error", errors.ErrorStack(err)))
```
kennytm: `zap.Error(err)` already included the error stack.

pkg/importer/job.go
```diff
@@ -33,24 +34,24 @@ func addJobs(jobCount int, jobChan chan struct{}) {
 func doInsert(table *table, db *sql.DB, count int) {
 	sqls, err := genRowDatas(table, count)
 	if err != nil {
-		log.Fatalf(errors.ErrorStack(err))
+		log.Fatal("genRowDatas", zap.String("error", errors.ErrorStack(err)))
```
kennytm: (ditto)

sync_diff_inspector/diff.go
```diff
@@ -414,7 +415,7 @@ func (df *Diff) Equal() (err error) {
 				return err
 			})
 			if err != nil {
-				log.Errorf("check %s.%s equal failed, error %v", table.Schema, table.Table, errors.ErrorStack(err))
+				log.Error("check failed", zap.String("table", dbutil.TableName(table.Schema, table.Table)), zap.String("error", errors.ErrorStack(err)))
```
kennytm: (ditto)

sync_diff_inspector/main.go
```diff
@@ -33,22 +34,21 @@ func main() {
 	case flag.ErrHelp:
 		os.Exit(0)
 	default:
-		log.Errorf("parse cmd flags err %s\n", errors.ErrorStack(err))
+		log.Error("parse cmd flags", zap.String("error", errors.ErrorStack(err)))
```
kennytm: (ditto)

sync_diff_inspector/main.go
```diff
@@ -33,22 +34,21 @@ func main() {
...
 		os.Exit(2)
 	}
 
 	if cfg.PrintVersion {
-		log.Infof("version: \n%s", utils.GetRawInfo("sync_diff_inspector"))
+		log.Info("sync_diff_inspector", zap.String("version", utils.GetRawInfo("sync_diff_inspector")))
```
kennytm: I think it is more appropriate to `fmt.Printf` this?

tidb-binlog/binlogctl/nodes.go
```diff
@@ -43,7 +44,7 @@ func queryNodesByKind(urls string, kind string) error {
 	}
 
 	for _, n := range nodes {
-		log.Infof("%s: %s", kind, formatNodeInfo(n))
+		log.Info("query node", zap.String("type", kind), zap.String("node", n.String()))
```
kennytm:
```suggestion
		log.Info("query node", zap.String("type", kind), zap.Stringer("node", n))
```

tidb-binlog/driver/example/mysql/mysql.go
```diff
@@ -276,7 +277,7 @@ func toSQL(binlog *pb.Binlog) ([]string, [][]interface{}) {
 		}
 
 	default:
-		log.Fatal("unknown type: ", binlog.GetType())
+		log.Fatal("unknown type", zap.String("type", binlog.GetType().String()))
```
kennytm:
```suggestion
		log.Fatal("unknown type", zap.Stringer("type", binlog.GetType()))
```

tidb-binlog/driver/example/print/print.go
```diff
@@ -47,7 +48,7 @@ func main() {
 	for {
 		select {
 		case msg := <-breader.Messages():
-			log.Info("recv: ", msg.Binlog.String())
+			log.Info("recv", zap.String("message", msg.Binlog.String()))
```
kennytm:
```suggestion
			log.Info("recv", zap.Stringer("message", msg.Binlog))
```

tidb-binlog/pump_client/client.go
```diff
@@ -232,7 +230,7 @@ func (c *PumpsClient) getPumpStatus(pctx context.Context) (revision int64, err e
 	}
 
 	for _, status := range nodesStatus {
-		log.Debugf("[pumps client] get pump %v from pd", status)
+		log.Debug("[pumps client] get pump from pd", zap.String("pump", status.String()))
```
kennytm:
```suggestion
		log.Debug("[pumps client] get pump from pd", zap.Stringer("pump", status))
```

tidb-binlog/pump_client/client.go
```diff
@@ -485,28 +484,28 @@ func (c *PumpsClient) watchStatus(revision int64) {
 				status := &node.Status{}
 				err := json.Unmarshal(ev.Kv.Value, &status)
 				if err != nil {
-					log.Errorf("[pumps client] unmarshal pump status %q failed", ev.Kv.Value)
+					log.Error("[pumps client] unmarshal pump status failed", zap.String("value", string(ev.Kv.Value)), zap.Error(err))
```
kennytm:
```suggestion
					log.Error("[pumps client] unmarshal pump status failed", zap.ByteString("value", ev.Kv.Value), zap.Error(err))
```

pkg/dbutil/common.go
WangXiangUSTC: in some place I use zap.Any, it seems ok,  is it necessary to change zap.Any to zap.Reflect?

kennytm: `zap.Any` will check if the value is any of the specialized type, and forwards to `zap.Reflect` when all fails. Calling `zap.Reflect` directly just skips the type check. 

If the input is not an `interface{}`, I think using `zap.Reflect` is better.

WangXiangUSTC:
@kennytm PTAL again

go.mod1
```diff
@@ -5,85 +5,33 @@ require (
...
-	github.com/soheilhy/cmux v0.1.4 // indirect
-	github.com/spaolacci/murmur3 v0.0.0-20180118202830-f09979ecbc72 // indirect
-	github.com/tmc/grpc-websocket-proxy v0.0.0-20171017195756-830351dc03c6 // indirect
-	github.com/twinj/uuid v1.0.0 // indirect
-	github.com/uber/jaeger-client-go v2.15.0+incompatible // indirect
-	github.com/uber/jaeger-lib v1.5.0 // indirect
-	github.com/ugorji/go v1.1.1 // indirect
-	github.com/unrolled/render v0.0.0-20180914162206-b9786414de4d // indirect
-	github.com/xiang90/probing v0.0.0-20160813154853-07dd2e8dfe18 // indirect
+	github.com/sirupsen/logrus v1.2.0
```
kennytm: Is `logrus` still used somewhere?

pkg/dbutil/common.go
```diff
@@ -199,7 +200,7 @@ func GetRandomValues(ctx context.Context, db *sql.DB, schemaName, table, column
 
 	query := fmt.Sprintf("SELECT %[1]s, COUNT(*) count FROM (SELECT %[1]s FROM %[2]s WHERE %[3]s ORDER BY RAND() LIMIT %[4]d)rand_tmp GROUP BY %[1]s ORDER BY %[1]s%[5]s",
 		escapeName(column), TableName(schemaName, table), limitRange, num, collation)
-	log.Debugf("get random values sql: %s, args: %v", query, limitArgs)
+	log.Debug("get random values", zap.String("sql", query), zap.String("args", fmt.Sprintf("%v", limitArgs)))
```
kennytm:
```suggestion
	log.Debug("get random values", zap.String("sql", query), zap.Reflect("args", limitArgs))
```

pkg/dbutil/common.go
```diff
@@ -363,7 +364,7 @@ func GetCRC32Checksum(ctx context.Context, db *sql.DB, schemaName, tableName str
 
 	query := fmt.Sprintf("SELECT BIT_XOR(CAST(CRC32(CONCAT_WS(',', %s, CONCAT(%s)))AS UNSIGNED)) AS checksum FROM %s WHERE %s;",
 		strings.Join(columnNames, ", "), strings.Join(columnIsNull, ", "), TableName(schemaName, tableName), limitRange)
-	log.Debugf("checksum sql: %s, args: %v", query, args)
+	log.Debug("checksum", zap.String("sql", query), zap.String("args", fmt.Sprintf("%v", args)))
```
kennytm:
```suggestion
	log.Debug("checksum", zap.String("sql", query), zap.Reflect("args", args))
```

pkg/dbutil/common.go
```diff
@@ -372,7 +373,7 @@ func GetCRC32Checksum(ctx context.Context, db *sql.DB, schemaName, tableName str
 	}
 	if !checksum.Valid {
 		// if don't have any data, the checksum will be `NULL`
-		log.Warnf("get empty checksum by query %s, args %v", query, args)
+		log.Warn("get empty checksum", zap.String("sql", query), zap.String("args", fmt.Sprintf("%v", args)))
```
kennytm:
```suggestion
		log.Warn("get empty checksum", zap.String("sql", query), zap.Reflect("args", args))
```

kennytm (commented):
Rest LGTM

kennytm (commented):
LGTM

pkg/diff/diff.go
```diff
@@ -501,9 +502,9 @@ func compareData(map1, map2 map[string]*dbutil.ColumnData, orderKeyCols []*model
 		}
 		equal = false
 		if data1.IsNull == data2.IsNull {
-			log.Errorf("find difference data in column %s, data1: %s, data2: %s", key, data1.Data, data2.Data)
+			log.Error("find difference data", zap.String("column", key), zap.Reflect("data1", map1), zap.Reflect("data2", map2))
```
csuzhangxc, line 505: log `map1`/`map2` rather than `data1.Data`/`data2.Data`?

WangXiangUSTC: old log only print the column's data, not convient for select the line for debug, so I modify this log

pkg/diff/diff.go
```diff
@@ -583,7 +584,7 @@ func getChunkRows(ctx context.Context, db *sql.DB, schema, table string, tableIn
 	query := fmt.Sprintf("SELECT /*!40001 SQL_NO_CACHE */ %s FROM `%s`.`%s` WHERE %s ORDER BY %s%s",
 		columns, schema, table, where, strings.Join(orderKeys, ","), collation)
 
-	log.Debugf("select data by sql %s, args: %v", query, args)
+	log.Debug("select data", zap.String("sql", query), zap.Any("args", args))
```
csuzhangxc: replace `zap.Any` with `zap.Reflect`?
same in _tidb-binlog/driver/example/mysql/mysql.go_ L91.

csuzhangxc (commented):
rest LGTM

pkg/diff/diff.go
WangXiangUSTC: https://godoc.org/go.uber.org/zap#Reflect
in the document, have a word `Outside tests, Any is always a better choice`, should we replace `Any` to `Reflect`? @kennytm

csuzhangxc: I think this doc can be exchanged to something like
```
Reflect is always a better choice
```
:grimacing:

WangXiangUSTC: `but it's relatively slow and allocation-heavy`, seems not suggest 
😬

kennytm: @WangXiangUSTC @csuzhangxc 

1. If the type is `interface{}`, `Any` is indeed better than `Reflect` since it could fallback to faster methods if it turns out to be a string/int/etc. However `args` is an `[]interface{}`.
2. When you know none of the predefined methods applies, `Reflect` is a better choice than `Any`, since `Any` will eventually give up and call `Reflect` anyway.
3. Every other method is a better choice than both `Any` and `Reflect`.
4. You could implement [ArrayMarshaler](https://godoc.org/go.uber.org/zap/zapcore#ArrayMarshaler) to use `zap.Array`, for instance.

csuzhangxc: yeh. the problem here is `args []interface{}`.

WangXiangUSTC: I do a bench mark test for []interface{}, reflect is a little better than any

BenchmarkAnyLog-4                 500000             15240 ns/op
BenchmarkReflectLog-4             500000             15520 ns/op

Already changed, PTAL again

csuzhangxc (approved):
LGTM

PR #221, opened 2019-03-19
pump client: support change select pump's strategy

WangXiangUSTC:
support change select pump's strategy
proposal: https://github.com/pingcap/tidb/issues/9201

create a new selector when change strategy

### Check List 

Tests 

 - Unit test

Code changes


tidb-binlog/pump_client/client.go
```diff
@@ -241,14 +242,19 @@ func (c *PumpsClient) getPumpStatus(pctx context.Context) (revision int64, err e
 
 // WriteBinlog writes binlog to a situable pump. Tips: will never return error for commit/rollback binlog.
 func (c *PumpsClient) WriteBinlog(binlog *pb.Binlog) error {
+	c.RLock()
+	pumpNum := len(c.Pumps.AvaliablePumps)
+	selector := c.Selector
```
IANTHEREAL, line 245: why putting it into a lock scope？

IANTHEREAL: Ok，I think you can put lock in selector, and provide an update function. it's better

IANTHEREAL:
Rest LGTM

IANTHEREAL:
@july2993 Please take a careful look

tidb-binlog/pump_client/client.go
```diff
@@ -416,13 +418,21 @@ func (c *PumpsClient) addPump(pump *PumpStatus, updateSelector bool) {
...
 	}
 
-	c.Pumps.Unlock()
+	c.Unlock()
+}
+
+// SetSelectStrategy sets the selector's strategy, strategy should be 'range' or 'hash' now.
+func (c *PumpsClient) SetSelectStrategy(strategy string) {
+	c.Lock()
+	c.Selector = NewSelector(strategy)
```
july2993, line 429: will you check which is really supported before call SetSelectStrategy? like setting `score` must has some problem now.
better return err for wrong strategy and don't fallback the default strategy.

july2993 (commented):
rest LGTM
- you should try avoid  global var tsMap, selectorLock which is shard by multi client.
- currently support `range` and `hash`, but i think no different for usage, maybe just confuse people, test more about `hash` if expose this to user.

tidb-binlog/pump_client/client.go
WangXiangUSTC: when change strategy will create a new selector. And selector is a member of pump client. so I think lock pump client seems better

WangXiangUSTC:
1. if not use a global tsMap, need set tsMap to new selector from old selector, seems it is a little tedious.
2. when user change the strategy, and at the moment also have write binlog request, will have two selector, and the tsMap should use same one. 
3. one tidb instance should only have one pump client, so it dosen't matter

july2993:
i know it it doesn't matter now, you should encapsulate such thinks in your `PumpsClient`,  but i am ok with this currently.

july2993 (commented):
LGTM

IANTHEREAL:
In addition to the poor packaging, LGTM

WangXiangUSTC:
OK, I will refactor these code later. @GregoryIan

PR #224, opened 2019-03-26
sync-diff-inspector: support checkpoint

WangXiangUSTC:
save sync-diff-inspector's checkpoint, and then can stop it and restart to continue check.

1. update chunk's struct
2. save chunk's info to downstream when check
3. when start sync-diff-inspector, will load checkpoint information
4. refine some code

### Check List 

Tests 

 - Unit test
 - Manual test (stop and then restart)

Related changes


pkg/dbutil/common.go
```diff
@@ -617,3 +624,18 @@ func ReplacePlaceholder(str string, args []string) string {
...
 	return fmt.Sprintf(newStr, utils.StringsToInterfaces(args)...)
 }
+
+// ExecSQLWithRetry executes sql with retry
+func ExecSQLWithRetry(ctx context.Context, db *sql.DB, sql string, args ...interface{}) (err error) {
+	ctx1, cancel := context.WithTimeout(ctx, DefaultTimeout)
+	defer cancel()
+
+	for i := 0; i < DefaultRetryTime; i++ {
+		_, err = db.ExecContext(ctx1, sql, args...)
```
csuzhangxc: Does all retry (10 times) share the total timeout (5 seconds)?

pkg/dbutil/common.go
```diff
@@ -617,3 +624,18 @@ func ReplacePlaceholder(str string, args []string) string {
...
+		if err == nil {
```
csuzhangxc, line 646: some errors will never success (like table already exists), doing retry have no meaning. Should we need to add something like `isRetryableError` or `isIgnorableError`?

pkg/diff/chunk.go
```diff
@@ -34,69 +35,83 @@ var (
...
+	UpperSymbol string `json:"upper-symbol"`
 }
 
-// chunkRange represents chunk range
-type chunkRange struct {
-	bounds []*bound
+// ChunkRange represents chunk range
+type ChunkRange struct {
+	ID     int      `json:"id"`
+	Bounds []*Bound `json:"Bounds"`
```
csuzhangxc:
```suggestion
	Bounds []*Bound `json:"bounds"`
```

pkg/diff/chunk.go
```diff
@@ -34,69 +35,83 @@ var (
...
 }
 
-// String returns the string of chunkRange, used for log.
-func (c *chunkRange) String() string {
+// String returns the string of ChunkRange, used for log.
+func (c *ChunkRange) String() string {
 	var s strings.Builder
-	s.WriteString("{")
-	for _, bound := range c.bounds {
+	s.WriteString("{ ID: ")
```
csuzhangxc: can we encode it to JSON directly?

pkg/dbutil/common.go
WangXiangUSTC: yes, do we need set timeout for every retry?

pkg/dbutil/common.go
```diff
@@ -35,6 +42,12 @@ const (
 
 	// ImplicitColID is ID implicit column in TiDB
 	ImplicitColID = -1
+
+	// DefaultRetryTime is the default retry time for execute sql
```
IANTHEREAL:
```suggestion
	// DefaultRetryTime is the default retry time to execute sql
```

pkg/dbutil/common.go
```diff
@@ -617,3 +630,85 @@ func ReplacePlaceholder(str string, args []string) string {
...
+		switch mysqlErr.Number {
+		// ER_LOCK_DEADLOCK can retry to commit while meet deadlock
+		case tmysql.ErrUnknown, gmysql.ER_LOCK_DEADLOCK, tmysql.ErrPDServerTimeout, tmysql.ErrTiKVServerTimeout, tmysql.ErrTiKVServerBusy, tmysql.ErrResolveLockTimeout, tmysql.ErrRegionUnavailable:
+			return true
+		default:
+			return false
+		}
+	}
+
+	return false
```
IANTHEREAL: return true？ if we're sure the errors what we should try, It could return false, but I think sort out the errors what we shouldn't try is easier

pkg/dbutil/common.go
```diff
@@ -617,3 +630,85 @@ func ReplacePlaceholder(str string, args []string) string {
 	newStr := strings.Replace(str, "?", "'%s'", -1)
 	return fmt.Sprintf(newStr, utils.StringsToInterfaces(args)...)
 }
+
+// ExecSQLWithRetry executes sql with retry
+func ExecSQLWithRetry(ctx context.Context, db *sql.DB, sql string, args ...interface{}) (err error) {
```
IANTHEREAL, line 638: should we add test cases for these function？

pkg/diff/util.go
```diff
@@ -81,25 +99,210 @@ func getColumnsFromIndex(index *model.IndexInfo, tableInfo *model.TableInfo) []*
...
+		log.Error("save chunk info failed", zap.Error(err))
+		return err
 	}
+	return nil
+}
 
-	for j := 0; j < num; j++ {
-		r := j + rand.Intn(total-j)
-		totalArray[j], totalArray[r] = totalArray[r], totalArray[j]
+func loadFromCheckPoint(ctx context.Context, db *sql.DB, schema, table, configHash string) (bool, error) {
```
IANTHEREAL: put `checkpoint` in a separated file

pkg/dbutil/common.go
csuzhangxc: It's your choice, :wink:

pkg/diff/chunk.go
```diff
@@ -34,69 +35,67 @@ var (
...
+	LowerSymbol string `json:"lower-symbol"`
+	Upper       string `json:"upper"`
+	UpperSymbol string `json:"upper-symbol"`
 }
 
-// chunkRange represents chunk range
-type chunkRange struct {
-	bounds []*bound
+// ChunkRange represents chunk range
+type ChunkRange struct {
```
GregoryIan: I think we should only create object as [POC](https://docs.google.com/document/d/1i43s9LBQEdIUkrdnAnZMa-69RhiG5VrJ8xcu3qWVRYE/edit?usp=sharing), there are two advantages
* Fit design document
* less code, we don't need code to assemble or transform the objects

pkg/diff/chunk.go
```diff
@@ -34,69 +35,67 @@ var (
...
-		s.WriteString(" ], ")
-	}
-	s.WriteString("}")
-
-	return s.String()
+// String returns the string of ChunkRange, used for log.
+func (c *ChunkRange) String() string {
+	chunkBytes, err := json.Marshal(c)
+	if err != nil {
+		log.Warn("get string for chunk", zap.Error(err))
```
IANTHEREAL:
```suggestion
		log.Warn("fail to encode chunk into string", zap.Error(err))
```

pkg/diff/util.go
```diff
@@ -24,6 +28,20 @@ import (
 	"go.uber.org/zap"
 )
 
+var (
+	successState = "success"
+
+	failedState = "failed"
+
+	errorState = "error"
```
IANTHEREAL: add some comments for these `state`? what's differences between `failedState` and `errorState`?

WangXiangUSTC: add comment

pkg/dbutil/common.go
WangXiangUSTC: done, these code same with DM, @csuzhangxc PTAL

pkg/dbutil/common.go
WangXiangUSTC: add test for ignoreError and isRetryableError

pkg/dbutil/common.go
IANTHEREAL: if same, keep them. But another question, can dm reuse these functions?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+	// for chunk: means meet error when check, don't know the chunk's data is equal or not equal
+	// for table: don't have this state
+	errorState = "error"
+
+	// for chunk: means this chunk is not in check
+	// for table: all the chunk in this table is not in check
+	notCheckedState = "not_checked"
+
+	// for chunk: means this chunk is checking
+	// for table: some chunks in this table is checking
```
IANTHEREAL: `// for table: don't have this state`？

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+	+--------+-------+-----------+-------------------+------------------+------------------+---------+----------------------------------+---------------------+
+	| schema | table | chunk_num | check_success_num | check_failed_num | check_ignore_num | state   | config_hash                      | update_time         |
+	+--------+-------+-----------+-------------------+------------------+------------------+---------+----------------------------------+---------------------+
+	| diff   | test  |       112 |               104 |                0 |                8 | success | 91f302052783672b01af3e2b0e7d66ff | 2019-03-26 12:42:11 |
+	+--------+-------+-----------+-------------------+------------------+------------------+---------+----------------------------------+---------------------+
+	*/
+	createSummaryTableSQL := "CREATE TABLE IF NOT EXISTS `sync_diff_inspector`.`table_summary`(`schema` varchar(30), `table` varchar(30), `chunk_num` int, `check_success_num` int, `check_failed_num` int, `check_ignore_num` int, `state` enum('not_checked', 'checking', 'success', 'failed') DEFAULT 'not_checked', `config_hash` varchar(50), `update_time` datetime, PRIMARY KEY(`schema`, `table`));"
```
GregoryIan: use `` ` `` to format sql

IANTHEREAL: add comment for `config_hash`?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+
+	/* example
+	mysql> select * from sync_diff_inspector.chunk where chunk_id = 2;;
+	+----------+-------------+--------+-------+---------------------------------+-------------+-----------+---------+---------------------+
+	| chunk_id | instance_id | schema | table | range                           |  checksum   | chunk_str | state   | update_time         |
+	+----------+-------------+--------+-------+---------------------------------+-------------+-----------+---------+---------------------+
+	|        2 | target-1    | diff   | test1 | (`a` >= ? AND `a` < ? AND TRUE) |  91f3020527 |  .....    | success | 2019-03-26 12:41:42 |
+	+----------+-------------+--------+-------+---------------------------------+-------------+-----------+---------+---------------------+
+	*/
+	createChunkTableSQL := "CREATE TABLE IF NOT EXISTS `sync_diff_inspector`.`chunk`(`chunk_id` int, `instance_id` varchar(30), `schema` varchar(30), `table` varchar(30), `range` varchar(100), `checksum` varchar(20), `chunk_str` text, `state` enum('not_checked', 'checking', 'success', 'failed', 'ignore', 'error') DEFAULT 'not_checked', update_time datetime, PRIMARY KEY(`chunk_id`, `instance_id`, `schema`, `table`));"
```
IANTHEREAL: how about ``PRIMARY KEY(`schema`, `table`, `instance_id`, `chunk_id`)``? we would use this index in `cleanCheckpointInfo` and `updateSummaryInfo `

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+	_, err = db.ExecContext(ctx, createChunkTableSQL)
+	if err != nil {
+		log.Info("create chunk table", zap.Error(err))
+		return errors.Trace(err)
+	}
+
+	return nil
+}
+
+func cleanCheckpointInfo(ctx context.Context, db *sql.DB, schema, table string) error {
```
IANTHEREAL: put them into one transaction?

pkg/dbutil/common.go
```diff
@@ -617,3 +629,84 @@ func ReplacePlaceholder(str string, args []string) string {
...
+	ctx1, cancel := context.WithTimeout(ctx, DefaultTimeout)
```
IANTHEREAL: I think it's better to set a timeout from caller

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+			return nil, err
+		}
+		chunks = append(chunks, chunk)
+	}
+
+	return chunks, nil
+}
+
+func updateSummaryInfo(ctx context.Context, db *sql.DB, instanceID, schema, table string) error {
+	ctx, cancel := context.WithTimeout(ctx, dbutil.DefaultTimeout)
```
IANTHEREAL: ditto, it's better to set timeout from caller

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+			successNum = num.Int64
+		case failedState, errorState:
+			failedNum += num.Int64
+		case ignoreState:
+			ignoreNum += num.Int64
+		case notCheckedState, checkingState:
+		}
+	}
+
+	log.Info("summary info", zap.String("instance_id", instanceID), zap.String("schema", schema), zap.String("table", table), zap.Int64("chunk num", total), zap.Int64("success num", successNum), zap.Int64("failed num", failedNum), zap.Int64("ignore num", ignoreNum))
```
IANTHEREAL, line 230: how is the performance of this log library?

pkg/dbutil/common.go
```diff
@@ -617,3 +629,84 @@ func ReplacePlaceholder(str string, args []string) string {
...
+		switch mysqlErr.Number {
+		// ER_LOCK_DEADLOCK can retry to commit while meet deadlock
+		case tmysql.ErrUnknown, gmysql.ER_LOCK_DEADLOCK, tmysql.ErrPDServerTimeout, tmysql.ErrTiKVServerTimeout, tmysql.ErrTiKVServerBusy, tmysql.ErrResolveLockTimeout, tmysql.ErrRegionUnavailable:
+			return true
+		default:
+			return false
+		}
+	}
+
+	return true
```
csuzhangxc: for `nil` (`!ok`), return `true` or `false` which is better?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+			state = successState
+		} else {
+			state = failedState
+		}
+	}
+
+	updateSQL := "UPDATE `sync_diff_inspector`.`table_summary` SET `chunk_num` = ?, `check_success_num` = ?, `check_failed_num` = ?, `check_ignore_num` = ?, `state` = ?, `update_time` = ? WHERE `schema` = ? AND `table` = ?"
+	err = dbutil.ExecSQLWithRetry(ctx, db, updateSQL, total, successNum, failedNum, ignoreNum, state, time.Now(), schema, table)
+	if err != nil {
+		return err
```
IANTHEREAL: we use `errors.Trace` in some places, can we unify them?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+	err := dbutil.ExecSQLWithRetry(ctx, db, sql, schema, table, notCheckedState, configHash, time.Now())
+	if err != nil {
+		log.Error("save summary info failed", zap.Error(err))
+		return err
+	}
+
+	return nil
+}
+
+func loadChunksInfo(ctx context.Context, db *sql.DB, instanceID, schema, table string) ([]*ChunkRange, error) {
```
IANTHEREAL:
```suggestion
func loadChunks(ctx context.Context, db *sql.DB, instanceID, schema, table string) ([]*ChunkRange, error) {
```

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+			}
+		}
+
+		return true, nil
+	}
+
+	return false, nil
+}
+
+func initSummaryInfo(ctx context.Context, db *sql.DB, schema, table string, configHash string) error {
```
IANTHEREAL:
```suggestion
func initTableSummary(ctx context.Context, db *sql.DB, schema, table string, configHash string) error {
```

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+	// for chunk: means this chunk is checking
+	// for table: some chunks in this table is checking
+	checkingState = "checking"
+
+	// for chunk: this chunk is ignored. if sample is not 100%, will ignore some chunk
+	// for table: don't have this state
+	ignoreState = "ignore"
+)
+
+func saveChunkInfo(ctx context.Context, db *sql.DB, chunkID int, instanceID, schema, table, checksum string, chunk *ChunkRange) error {
```
IANTHEREAL:
```suggestion
func saveChunk(ctx context.Context, db *sql.DB, chunkID int, instanceID, schema, table, checksum string, chunk *ChunkRange) error {
```

pkg/dbutil/common.go
```diff
@@ -617,3 +629,84 @@ func ReplacePlaceholder(str string, args []string) string {
...
+	// TODO: now only ignore some ddl error, add some dml error later
+	if ignoreDDLError(err) {
+		return true
+	}
+
+	return false
+}
+
+func ignoreDDLError(err error) bool {
+	err = originError(err)
```
csuzhangxc: just use `err = errors.Cause(err)`?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+
+		if cfgHash.Valid {
+			if configHash != cfgHash.String {
+				return false, nil
+			}
+		}
+
+		if state.Valid {
+			// if state is not checked, the chunk info maybe not exists, so just return false
+			if state.String == successState || state.String == notCheckedState {
```
GregoryIan: also `state.String == successState`

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+	sql := "REPLACE INTO `sync_diff_inspector`.`chunk` VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?);"
+	err = dbutil.ExecSQLWithRetry(ctx, db, sql, chunkID, instanceID, schema, table, chunk.Where, checksum, string(chunkBytes), chunk.State, time.Now())
+	if err != nil {
+		log.Error("save chunk info failed", zap.Error(err))
+		return err
+	}
+	return nil
+}
+
+func loadFromCheckPoint(ctx context.Context, db *sql.DB, schema, table, configHash string) (bool, error) {
```
IANTHEREAL: what's meaning of this function?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+
+		chunkStr := fields["chunk_str"].Data
+		chunk := new(ChunkRange)
+		err := json.Unmarshal(chunkStr, &chunk)
+		if err != nil {
+			return nil, err
+		}
+		chunks = append(chunks, chunk)
+	}
+
```
IANTHEREAL, line 133: check `rows.Err()`?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+		case successState:
+			successNum = num.Int64
+		case failedState, errorState:
+			failedNum += num.Int64
+		case ignoreState:
+			ignoreNum += num.Int64
+		case notCheckedState, checkingState:
+		}
+	}
+
```
IANTHEREAL, line 204: check `rows.Err()`?

pkg/dbutil/common.go
```diff
@@ -617,3 +629,84 @@ func ReplacePlaceholder(str string, args []string) string {
...
+		return false
+	}
+
+	errCode := terror.ErrCode(mysqlErr.Number)
+	switch errCode {
+	case infoschema.ErrDatabaseExists.Code(), infoschema.ErrDatabaseDropExists.Code(),
+		infoschema.ErrTableExists.Code(), infoschema.ErrTableDropExists.Code(),
+		infoschema.ErrColumnExists.Code(), infoschema.ErrIndexExists.Code():
+		return true
+	case tmysql.ErrDupKeyName:
```
csuzhangxc: is any type cast needed? I don't confirm this in DM is correct.

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+	notCheckedState = "not_checked"
+
+	// for chunk: means this chunk is checking
+	// for table: some chunks in this table is checking
+	checkingState = "checking"
+
+	// for chunk: this chunk is ignored. if sample is not 100%, will ignore some chunk
+	// for table: don't have this state
+	ignoreState = "ignore"
+)
```
IANTHEREAL, line 59: if we add some interface for checkpoint, it would fine for unit test, cuz we can mock it

pkg/diff/chunk.go
```diff
@@ -516,19 +516,18 @@ func getSplitFields(table *model.TableInfo, splitFields []string) ([]*model.Colu
...
-// CheckJob is the struct of job for check
-type CheckJob struct {
-	Schema string
-	Table  string
-	Where  string
-	Args   []string
-}
+// GetChunks gets some chunks.
+func GetChunks(ctx context.Context, table *TableInstance, splitFields, limits string, chunkSize int, collation string, useTiDBStatsInfo bool) (chunks []*ChunkRange, fromCheckpoint bool, err error) {
+	chunks, err = loadChunksInfo(ctx, table.Conn, table.InstanceID, table.Schema, table.Table)
```
IANTHEREAL: what if fail to save some chunks?

pkg/diff/diff.go
```diff
@@ -32,78 +36,98 @@ import (
...
+}
+
+func (t *TableDiff) setConfigHash() error {
+	jsonBytes, err := json.Marshal(t)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
+	t.configHash = fmt.Sprintf("%x", md5.Sum(jsonBytes))
+	log.Debug("table diff config", zap.ByteString("config", jsonBytes), zap.String("hash", t.configHash))
```
IANTHEREAL:
```suggestion
	log.Debug("sync-diff-inspector config", zap.ByteString("config", jsonBytes), zap.String("hash", t.configHash))
```

pkg/diff/diff.go
```diff
@@ -184,58 +250,55 @@ func (t *TableDiff) CheckTableData(ctx context.Context) (bool, error) {
...
+	chunks, fromCheckpoint, err := GetChunks(ctx, table, t.Fields, t.Range, t.ChunkSize, t.Collation, useTiDB)
 	if err != nil {
 		return false, errors.Trace(err)
 	}
-
-	checkNums := len(allJobs) * t.Sample / 100
-	checkNumArr := getRandomN(len(allJobs), checkNums)
-	log.Info("check jobs", zap.Int("total job num", len(allJobs)), zap.Int("check job num", len(checkNumArr)))
-	if checkNums == 0 {
+	if len(chunks) == 0 {
```
IANTHEREAL, line 221: what if just downstream doesn't have any data, but upstream has some data?

pkg/diff/diff.go
```diff
@@ -259,63 +322,121 @@ func (t *TableDiff) getSourceTableChecksum(ctx context.Context, job *CheckJob) (
...
+	if filterByRand {
+		rand.Seed(time.Now().UnixNano())
+		r := rand.Intn(100)
+		if r > t.Sample {
+			chunk.State = ignoreState
+			return true, nil
 		}
+	}
+
+	chunk.State = checkingState
```
IANTHEREAL, line 400: I always think it's useless

pkg/diff/diff.go
```diff
@@ -259,63 +322,121 @@ func (t *TableDiff) getSourceTableChecksum(ctx context.Context, job *CheckJob) (
...
 		if err != nil {
 			return false, errors.Trace(err)
 		}
-
-		// if equal is false, we continue check data, we should find all the different data just run once
-		if !eq {
-			equal = false
+		if equal {
+			return true, nil
 		}
```
GregoryIan: else output some log？

pkg/diff/diff.go
```diff
@@ -125,11 +149,53 @@ func (t *TableDiff) Equal(ctx context.Context, writeFixSQL func(string) error) (
...
 	}
 
-	t.sqlCh <- "end"
+	stopWriteSqlsCh <- true
```
IANTHEREAL, line 151: make sure all fixed sqls have written down

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+
+func saveChunk(ctx context.Context, db *sql.DB, chunkID int, instanceID, schema, table, checksum string, chunk *ChunkRange) error {
+	chunkBytes, err := json.Marshal(chunk)
+	if err != nil {
+		return err
+	}
+
+	sql := "REPLACE INTO `sync_diff_inspector`.`chunk` VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?);"
```
csuzhangxc: use variables to hold schema/table names to avoid typo?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,256 @@
...
+
+	/* example
+	mysql> select * from sync_diff_inspector.chunk where chunk_id = 2;;
+	+----------+-------------+--------+-------+---------------------------------+-------------+-----------+---------+---------------------+
+	| chunk_id | instance_id | schema | table | range                           |  checksum   | chunk_str | state   | update_time         |
+	+----------+-------------+--------+-------+---------------------------------+-------------+-----------+---------+---------------------+
+	|        2 | target-1    | diff   | test1 | (`a` >= ? AND `a` < ? AND TRUE) |  91f3020527 |  .....    | success | 2019-03-26 12:41:42 |
+	+----------+-------------+--------+-------+---------------------------------+-------------+-----------+---------+---------------------+
+	*/
+	createChunkTableSQL := "CREATE TABLE IF NOT EXISTS `sync_diff_inspector`.`chunk`(`chunk_id` int, `instance_id` varchar(30), `schema` varchar(30), `table` varchar(30), `range` varchar(100), `checksum` varchar(20), `chunk_str` text, `state` enum('not_checked', 'checking', 'success', 'failed', 'ignore', 'error') DEFAULT 'not_checked', update_time datetime, PRIMARY KEY(`chunk_id`, `instance_id`, `schema`, `table`));"
```
csuzhangxc: for `update_time`, set it to `ON UPDATE CURRENT_TIMESTAMP`?

pkg/diff/checkpoint.go
WangXiangUSTC: if the mode is `online check`, don't need this state. But for normal mode, need use this state.
I will update this comment when support online check

pkg/diff/diff.go
WangXiangUSTC: this already do in chunk splitter，this should never happened, I add a warn log

pkg/diff/diff.go
WangXiangUSTC: some sql may be blocked, we can select table by `state = checking` and check blocked in which chunk

pkg/dbutil/common.go
WangXiangUSTC: I update to use ddl.ErrDupKeyName.Code()

pkg/diff/checkpoint.go
WangXiangUSTC: I haven't do any test , I will ask the author of pingcap/log

pkg/diff/chunk.go
WangXiangUSTC: will recheck these chunks

WangXiangUSTC:
@GregoryIan @csuzhangxc PTAL again

pkg/diff/checkpoint.go
WangXiangUSTC: https://github.com/uber-go/zap#performance

pkg/dbutil/common.go
```diff
@@ -617,3 +633,116 @@ func ReplacePlaceholder(str string, args []string) string {
 	newStr := strings.Replace(str, "?", "'%s'", -1)
 	return fmt.Sprintf(newStr, utils.StringsToInterfaces(args)...)
 }
+
+// ExecSQLWithRetry executes sql with retry
+func ExecSQLWithRetry(ctx context.Context, db *sql.DB, timeout time.Duration, sql string, args ...interface{}) (err error) {
+	ctx1, cancel := context.WithTimeout(ctx, timeout)
```
IANTHEREAL: remove all this code, we should let caller to control timeout

pkg/diff/checkpoint.go
IANTHEREAL: if not online check, we must know it's checking, right? we can make thing simple

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,394 @@
...
+			return nil, err
+		}
+		chunks = append(chunks, chunk)
+	}
+
+	return chunks, errors.Trace(rows.Err())
+}
+
+// getSummary returns a table's total chunk num, check success chunk num, check failed chunk num, check ignore chunk num and the state
+func getSummary(ctx context.Context, db *sql.DB, timeout time.Duration, schema, table string) (int64, int64, int64, int64, string, error) {
```
IANTHEREAL:
```suggestion
func getSummary(ctx context.Context, db *sql.DB, timeout time.Duration, schema, table string) (total int64, success int64, failed int64, ignored int64, state string, err error) {
```

IANTHEREAL:
```suggestion
func getTableSummary(ctx context.Context, db *sql.DB, timeout time.Duration, schema, table string) (int64, int64, int64, int64, string, error) {
```

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,394 @@
...
+	}
+	if rows.Err() != nil {
+		return 0, 0, 0, 0, "", errors.Trace(rows.Err())
+	}
+
+	return 0, 0, 0, 0, "", errors.NotFoundf("schema %s, table %s summary info", schema, table)
+}
+
+// getSummaryFromChunk get the table's summary info from `chunk` table
+func getSummaryFromChunk(ctx context.Context, db *sql.DB, timeout time.Duration, instanceID, schema, table string) (total, successNum, failedNum, ignoreNum int64, err error) {
```
IANTHEREAL:
```suggestion
func getChunkSummary(ctx context.Context, db *sql.DB, timeout time.Duration, instanceID, schema, table string) (total, successNum, failedNum, ignoreNum int64, err error) {
```

pkg/diff/checkpoint.go
WangXiangUSTC: ok, the comment is updated

pkg/dbutil/common.go
```diff
@@ -617,3 +633,110 @@ func ReplacePlaceholder(str string, args []string) string {
...
+		_, err = db.ExecContext(ctx, sql, args...)
+		if err == nil {
+			takeDuration := time.Since(startTime)
+			if takeDuration > SlowWarnLog {
+				log.Warn("exec sql slow", zap.String("sql", sql), zap.Reflect("args", args), zap.Duration("take", takeDuration))
+			}
+			return nil
+		}
+
+		if ignoreError(err) {
```
csuzhangxc, line 650: move this block before L642, and reset err to `nil` (`err = nil`) in it, so we can reuse the `SlowWarnLog` mechanism？

pkg/dbutil/common.go
```diff
@@ -617,3 +633,110 @@ func ReplacePlaceholder(str string, args []string) string {
...
+			log.Debug("ignore execute sql error", zap.Error(err))
+			return nil
+		}
+
+		if !isRetryableError(err) {
+			return errors.Trace(err)
+		}
```
csuzhangxc, line 657: log a warn before retry?

pkg/dbutil/common.go
WangXiangUSTC: updated in 06ee87d, print slow log before judge error. do you think it's ok ?

csuzhangxc: fine

pkg/dbutil/common.go
```diff
@@ -617,3 +633,113 @@ func ReplacePlaceholder(str string, args []string) string {
...
+
+		log.Warn("exe sql failed, will try again", zap.String("sql", sql), zap.Reflect("args", args), zap.Error(err))
+		time.Sleep(10 * time.Millisecond)
```
csuzhangxc: 1. do we need to check `ctx.Done` here (maybe with another `time.After` `case` for `select`)?

2. if put this block at the end of the `for` loop, maybe need to check `i == DefaultRetryTime-1`, or we can put it at the begining of the `for` loop.

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,373 @@
...
+	var totalNum, successNum, failedNum, ignoreNum sql.NullInt64
+	var stateStr sql.NullString
+	for rows.Next() {
+		err1 := rows.Scan(&totalNum, &successNum, &failedNum, &ignoreNum, &stateStr)
+		if err1 != nil {
+			return 0, 0, 0, 0, "", errors.Trace(err1)
+		}
+
+		if !totalNum.Valid || !successNum.Valid || !failedNum.Valid || !ignoreNum.Valid || !stateStr.Valid {
+			return 0, 0, 0, 0, "", errors.New("get summary failed, value is valid")
```
csuzhangxc:
```suggestion
			return 0, 0, 0, 0, "", errors.New("get summary failed, some values are invalid")
```
?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,373 @@
...
+	var chunkState sql.NullString
+	var num sql.NullInt64
+	for rows.Next() {
+		err1 := rows.Scan(&chunkState, &num)
+		if err1 != nil {
+			return 0, 0, 0, 0, errors.Trace(err1)
+		}
+
+		if !chunkState.Valid || !num.Valid {
+			continue
```
csuzhangxc: do not need to return an error?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,373 @@
...
+		}
+
+		total += num.Int64
+		switch chunkState.String {
+		case successState:
+			successNum = num.Int64
```
csuzhangxc:
```suggestion
			successNum += num.Int64
```
?

pkg/diff/checkpoint.go
```diff
@@ -0,0 +1,373 @@
...
+		case ignoreState:
+			ignoreNum += num.Int64
+		case notCheckedState, checkingState:
+		}
+	}
+	if rows.Err() != nil {
+		return 0, 0, 0, 0, errors.Trace(rows.Err())
+	}
+
+	return
```
csuzhangxc, line 209: no NotFound error at here?

pkg/diff/chunk.go
```diff
@@ -541,31 +543,35 @@ func GenerateCheckJob(table *TableInstance, splitFields, limits string, chunkSiz
...
+		chunk.ID = i
+		chunk.Where = fmt.Sprintf("(%s AND %s)", conditions, limits)
+		chunk.Args = args
+		chunk.State = notCheckedState
 
-	for _, chunk := range chunks {
-		conditions, args := chunk.toString(mode, collation)
-		where := fmt.Sprintf("(%s AND %s)", conditions, limits)
+		ctx2, cancel2 := context.WithTimeout(ctx, dbutil.DefaultTimeout)
+		defer cancel2()
```
csuzhangxc: use `defer` in the `for` loop is not a good choice.

pkg/diff/diff.go
```diff
@@ -32,81 +36,101 @@ import (
...
-	CheckThreadCount int
+	CheckThreadCount int `json:"-"`
 
 	// set true if target-db and source-db all support tidb implicit column "_tidb_rowid"
-	UseRowID bool
+	UseRowID bool `json:"use-rowid"`
 
 	// set false if want to comapre the data directly
-	UseChecksum bool
+	UseChecksum bool `json:"-"`
```
csuzhangxc, line 81: no need to save this config item (and `IgnoreStructCheck`/`IgnoreDataCheck`)? if the user changes it, what is the behavior?

pkg/diff/diff.go
```diff
@@ -32,81 +36,101 @@ import (
...
 
 	t.sqlCh = make(chan string)
-	t.wg.Add(1)
-	go func() {
-		t.WriteSqls(ctx, writeFixSQL)
-		t.wg.Done()
-	}()
 
-	err := t.getTableInfo(ctx)
+	t.wg.Add(2)
```
csuzhangxc: I think put `wg.Add` and `wg.Done` as nearer as possible is better.
And extract `wg.Add`/`wg.Done` from concrete `func` is better.
The principle is split goroutine control from the business logic.

but now, you can just keep it.

pkg/diff/diff.go
```diff
@@ -187,58 +256,56 @@ func (t *TableDiff) CheckTableData(ctx context.Context) (bool, error) {
...
+	defer func() {
+		for _, ch := range checkWorkerCh {
+			close(ch)
+		}
+	}()
+
+	var checkedNum int
+	go func() {
+		for _, chunk := range chunks {
+			checkWorkerCh[chunk.ID%t.CheckThreadCount] <- chunk
```
csuzhangxc: it may be blocked if current `checkWorkerCh[chunk.ID%t.CheckThreadCount]` chan is too slow even other channels are empty?

pkg/diff/chunk.go
IANTHEREAL: would populate chunk again? I understand the progress is: populate chunk/load chunk -> save chunk -> check chunk

pkg/diff/diff.go
```diff
@@ -187,58 +256,56 @@ func (t *TableDiff) CheckTableData(ctx context.Context) (bool, error) {
...
-			num++
+			checkedNum++
 			if !eq {
 				equal = false
 			}
-			if num == t.CheckThreadCount {
+			if len(chunks) == checkedNum {
 				break CheckResult
 			}
 		case <-ctx.Done():
```
csuzhangxc, line 272: is it possible we canceled and `defer` in L285 called, but still trying to send chunk in L294?

pkg/diff/diff.go
WangXiangUSTC: only when some config are changed and need split chunk again should influence in the config hash

pkg/diff/diff.go
WangXiangUSTC: just make the code simple

pkg/diff/diff.go
WangXiangUSTC: you means should let it exit gracefully?

WangXiangUSTC:
@GregoryIan @csuzhangxc PTAL again

pkg/diff/chunk.go
WangXiangUSTC: in fact the progress is :  populate chunk -> save chunk -> check chunk -> update chunk

WangXiangUSTC: if meet some error when save chunks ,the table's state will be notChecked, and when diff restart, will clean checkpoint and  populate chunk again

IANTHEREAL:
LGTM

pkg/diff/diff.go
csuzhangxc: yeh, if we send to a closed, a `panic` will occur.
at least, we should wait the goroutine in L249 returned before close channels.

pkg/diff/diff.go
```diff
@@ -262,68 +337,128 @@ func (t *TableDiff) getSourceTableChecksum(ctx context.Context, job *CheckJob) (
...
-			if sourceChecksum == targetChecksum {
-				log.Info("checksum is equal", zap.String("table", dbutil.TableName(job.Schema, job.Table)), zap.String("where", job.Where), zap.Reflect("args", job.Args), zap.Int64("checksum", sourceChecksum))
+			if chunk.State == successState || chunk.State == ignoreState {
+				resultCh <- true
 				continue
 			}
-
-			log.Warn("checksum is not equal", zap.String("table", dbutil.TableName(job.Schema, job.Table)), zap.String("where", job.Where), zap.Reflect("args", job.Args), zap.Int64("source checksum", sourceChecksum), zap.Int64("target checksum", targetChecksum))
+			eq, err := t.checkChunkDataEqual(ctx, filterByRand, chunk)
+			if err != nil {
```
csuzhangxc, line 352: whether needing to log some info about the `chunk` when `err !=nil` or `!eq`?
or should we need to find out it from the checkpoint table?

pkg/diff/diff.go
WangXiangUSTC: updated in https://github.com/pingcap/tidb-tools/pull/224/commits/1c68b9a5a449ca5876b62b4a3cdefee74860d20a

pkg/diff/diff.go
```diff
@@ -262,68 +337,128 @@ func (t *TableDiff) getSourceTableChecksum(ctx context.Context, job *CheckJob) (
...
+
+	log.Warn("checksum is not equal", zap.String("table", dbutil.TableName(t.TargetTable.Schema, t.TargetTable.Table)), zap.String("where", chunk.Where), zap.Reflect("args", chunk.Args), zap.Int64("source checksum", sourceChecksum), zap.Int64("target checksum", targetChecksum))
+
+	return false, nil
+}
+
+func (t *TableDiff) compareRows(ctx context.Context, chunk *ChunkRange) (bool, error) {
+	sourceRows := make(map[string][]map[string]*dbutil.ColumnData)
+	for i, sourceTable := range t.SourceTables {
+		rows, _, err := getChunkRows(ctx, sourceTable.Conn, sourceTable.Schema, sourceTable.Table, sourceTable.info, chunk.Where, utils.StringsToInterfaces(chunk.Args), utils.SliceToMap(t.IgnoreColumns), t.Collation)
```
csuzhangxc: pre-calculate `utils.StringsToInterfaces(chunk.Args), utils.SliceToMap(t.IgnoreColumns)` before the `for` loop, reuse them both for source tables and target table?

pkg/diff/diff.go
```diff
@@ -431,23 +566,78 @@ func (t *TableDiff) compareRows(sourceRows map[string][]map[string]*dbutil.Colum
...
+					return
+				}
+
+				err := writeFixSQL(fmt.Sprintf("%s\n", dml))
+				if err != nil {
+					log.Error("write sql failed", zap.String("sql", dml), zap.Error(err))
+				}
+				t.wg.Done()
+			case <-stopWriteCh:
+				stop = true
```
csuzhangxc, line 596: only mark `stop = true` rather than `return`, so we can get all dml from `t.sqlCh`?

sync_diff_inspector/config.toml
```diff
@@ -24,6 +24,9 @@ use-checksum = true
 # set true if just want compare data by checksum, will skip select data when checksum is not equal. 
 only-use-checksum = false
 
+# set true will continue check from the latest checkpoint
+use-checkpoint = true
```
csuzhangxc, line 28: default `true` or `false`, which is better in this sample config.

sync_diff_inspector/config.go
```diff
@@ -229,6 +232,7 @@ func NewConfig() *Config {
 	fs.BoolVar(&cfg.PrintVersion, "V", false, "print version of sync_diff_inspector")
 	fs.BoolVar(&cfg.IgnoreDataCheck, "ignore-data-check", false, "ignore check table's data")
 	fs.BoolVar(&cfg.IgnoreStructCheck, "ignore-struct-check", false, "ignore check table's struct")
+	fs.BoolVar(&cfg.UseCheckpoint, "use-checkpoint", true, "set true will continue check from the latest checkpoint")
```
csuzhangxc, line 235: default to be `true`?

pkg/diff/diff.go
WangXiangUSTC: yes, if can read data from t.sqlCh, will not run default

sync_diff_inspector/config.toml
WangXiangUSTC: for `use-checkpoint`, I think default value is true is better

sync_diff_inspector/config.go
WangXiangUSTC: ditto

pkg/diff/diff.go
WangXiangUSTC: updated in https://github.com/pingcap/tidb-tools/pull/224/commits/5f204554c963e6817d528bc73f2d9a2fb7f105d6

pkg/diff/diff.go
WangXiangUSTC: updated in 5f20455

csuzhangxc (approved):
LGTM

PR #230, opened 2019-04-17
pump_client: Use `CommitTS % numberOfPumps` instead of hashing

suzaku:

As illustrated in this benchmark: https://play.golang.org/p/K7ttWMTZTQG
Encoding the TS directly into a byte array is about twice faster than
converting it to a string.

Use encoding/binary to turn the TS into a byte array.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


tidb-binlog/pump_client/selector_test.go
```diff
@@ -0,0 +1,28 @@
...
+	for i := 0; i < nTS; i++ {
+		counter[hashTs(int64(i))%nPumps]++
+	}
+	avg := nTS / 11
+	for i := 0; i < nPumps; i++ {
+		diff := counter[i] - avg
+		if diff < 0 {
+			diff = -diff
+		}
+		c.Assert(diff < 300, IsTrue)
```
kennytm:
```suggestion
		c.Assert(diff, Less, 300)
```

tidb-binlog/pump_client/selector.go
```diff
@@ -246,7 +246,9 @@ func NewSelector(strategy string) PumpSelector {
 
 func hashTs(ts int64) int {
 	h := fnv.New32a()
-	h.Write([]byte(strconv.FormatInt(ts, 10)))
+	b := make([]byte, binary.MaxVarintLen64)
+	n := binary.PutVarint(b, ts)
+	h.Write(b[:n])
```
kennytm: Why just write the fixed 8 bytes?

```go
	b := make([]byte, 8)
	binary.LittleEndian.PutUint64(b, uint64(ts))
	h.Write(b)
```

```
$ go test 1_test.go -test.bench .
goos: darwin
goarch: amd64
BenchmarkStrFirst-4   	10000000	       123 ns/op
BenchmarkTiBin-4      	20000000	        55.6 ns/op
BenchmarkTiFixed-4    	30000000	        48.8 ns/op
PASS
ok  	command-line-arguments	4.085s
```

suzaku: Cool, using all the 8 bytes is faster.

tidb-binlog/pump_client/selector.go
```diff
@@ -246,7 +246,9 @@ func NewSelector(strategy string) PumpSelector {
 
 func hashTs(ts int64) int {
```
july2993: i think we don't need this, just use ts % nPump?

suzaku:
@kennytm @july2993 PTAL

tidb-binlog/pump_client/selector_test.go
```diff
@@ -0,0 +1,28 @@
...
+var _ = Suite(&testSelectorSuite{})
+
+func (t *testSelectorSuite) TestHashTS(c *C) {
+	const (
+		nPumps = 11
+		nTS    = 1024
+	)
+	counter := [nPumps]int{}
+	for i := 0; i < nTS; i++ {
+		counter[hashTs(int64(i))%nPumps]++
```
kennytm: You need to cast the result of `hashTs` to an unsigned integer before taking modulo.

Currently `hashTs` may return a negative number, and e.g. `(-50) % 11 == -6`. The range of counter is thus -10 to 10, where the "0" slot is double-counted, i.e. the ideal distribution is like:

```
-10 = 4.5%
 -9 = 4.5%
...
 -1 = 4.5%
  0 = 9.1% <---
  1 = 4.5%
...
  9 = 4.5%
 10 = 4.5%
```

If we remove the double-counting, the deviation from average should be much less than 300.

tidb-binlog/pump_client/selector.go
suzaku: Yes, I think the `hash` strategy doesn't make sense if the only input is `TS`.

tidb-binlog/pump_client/selector_test.go
suzaku: Yes, the problem is that in 32-bit environment (eg. go playground), `int(h.Sum32())` may be negative.

kennytm: You could do `uint(hashTs) % nPumps`

tidb-binlog/pump_client/selector_test.go
```diff
@@ -0,0 +1,28 @@
...
+func (t *testSelectorSuite) TestHashTS(c *C) {
+	const (
+		nPumps = 11
+		nTS    = 1024
+	)
+	counter := [nPumps]int{}
+	for i := 0; i < nTS; i++ {
+		counter[uint(hashTs(int64(i)))%nPumps]++
+	}
+	avg := nTS / 11
```
kennytm:
```suggestion
	avg := nTS / nPumps
```

tidb-binlog/pump_client/selector_test.go
```diff
@@ -0,0 +1,28 @@
...
+	for i := 0; i < nPumps; i++ {
+		diff := counter[i] - avg
+		if diff < 0 {
+			diff = -diff
+		}
+		c.Assert(diff, Less, 300)
```
kennytm:
```suggestion
		c.Assert(diff, Less, 9)
```

nTS / 11 = 93 so with 300 this test will always pass, making the test useless.

"9" is the standard deviation of the binomial distribution B(1024, 1/11). If you want a tight bound you could use "5" instead of "9".

kennytm (commented):
Rest LGTM

tidb-binlog/pump_client/selector.go
WangXiangUSTC: I agree, can simply the hashTs function

suzaku: I mean we can just do without the `hash` strategy, `ts % nPump` is at best the `range` strategy, right?

suzaku:
@july2993 @kennytm  PTAL

WangXiangUSTC:
LGTM

tidb-binlog/pump_client/selector.go
```diff
@@ -98,7 +96,7 @@ func (h *HashSelector) Select(binlog *pb.Binlog, retryTime int) *PumpStatus {
 		return nil
 	}
 
-	pump := h.Pumps[(hashTs(binlog.StartTs)+int(retryTime))%len(h.Pumps)]
+	pump := h.Pumps[(int(binlog.StartTs)+retryTime)%len(h.Pumps)]
```
kennytm: Will this produce a negative number ?

suzaku: Updated

kennytm (approved):
LGTM

PR #236, opened 2019-05-02
column-mapping: set schemaID/tableID to 0 if numeric suffix is missing

kennytm:

(TOOL-1134) Support mixing names with and without numeric suffix.

Added an optional *separator* argument to column mapping. The suffix is extracted as *prefix* + *separator* + *suffix*.

If the *name* equals to *prefix* (without the separator), we treat this to mean using a zero numeric suffix, i.e. if `prefix == "money", separator == "_"`:

* `money` → 0
* `money_` → error
* `money_5` → 5
* `money_6` → 6
* `money_7` → 7
* `money_money_money` → error
* `mon` → error
* `not_money` → error

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/column-mapping/column.go
+++ b/pkg/column-mapping/column.go
@@ -120,6 +120,14 @@ func (r *Rule) Valid() error {
 	return nil
 }
 
+// Adjust normalizes the rule into an easier-to-process form, e.g. filling in
+// optional arguments with the default values.
+func (r *Rule) Adjust() {
+	if r.Expression == PartitionID && len(r.Arguments) == 3 {
+		r.Arguments = append(r.Arguments, "")
+	}
+}
+
 // check source and target position
 func (r *Rule) adjustColumnPosition(source, target int) (int, int, error) {
 	// if not found target, ignore it
@@ -182,9 +190,7 @@ func (m *Mapping) addOrUpdateRule(rule *Rule, isUpdate bool) error {
 	if !m.caseSensitive {
 		rule.ToLower()
 	}
-	if rule.Expression == PartitionID && len(rule.Arguments) == 3 {
-		rule.Arguments = append(rule.Arguments, "")
-	}
+	rule.Adjust()
 
 	m.resetCache()
 	err = m.Insert(rule.PatternSchema, rule.PatternTable, rule, isUpdate)
--- a/pkg/column-mapping/column.go
+++ b/pkg/column-mapping/column.go
@@ -506,7 +506,13 @@ func computePartitionID(schema, table string, rule *Rule) (instanceID int64, sch
 }
 
 func computeID(name string, prefix string, bitSize int, shiftCount uint) (int64, error) {
-	if len(prefix) >= len(name) || prefix != name[:len(prefix)] {
+	switch {
+	case len(prefix) >= len(name):
+		if prefix[:len(name)] == name {
+			return 0, nil
+		}
+		fallthrough
+	case prefix != name[:len(prefix)]:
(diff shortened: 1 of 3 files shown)


csuzhangxc:
@GregoryIan @ericsyh Have we decided to support this?

if so, `money`, `money_` and `mon` are all mapped to `0`, we need to underline it in the doc to avoid misuse.

kennytm:
We could also agree on a list of acceptable separators so that `money` and `money_` works but not `mo`.

IANTHEREAL:
mainly ambiguity problem, I think we can choose one in two way

* define prefix and separator, like prefix is `money` and separator is `_`,  it also means `money`, and `money_` are mapped to `0`
* only define prefix, like prefix is `money`,  `money` is mapped to `0` and ``money_` is invalid

The reason for this is based on the fact that we should provide a definition that is as simple and nature as possible, not easy to make mistakes. how about you? @kennytm @csuzhangxc

kennytm:
The second option seems better. I suggest the following changes:

1. We extend the arguments to 4 elements

    ```yaml
    arguments: [123, 'schemaprefix', 'tableprefix', '_']
    ```

    If the last argument is missing, it defaults to `''`

2. Schema name must be `schemaprefix` or `schemaprefix + sep + schemaID`
3. Table name must be `tableprefix` or `tableprefix + sep + tableID`

IANTHEREAL:
it's fine, and we still allow 3 elements arguments which implies `sep` is `""`

kennytm:
PTAL @csuzhangxc @GregoryIan @ericsyh

pkg/column-mapping/column.go
```diff
@@ -490,22 +495,33 @@ func computePartitionID(schema, table string, rule *Rule) (instanceID int64, sch
 		instanceID = int64(instanceIDUnsign << shiftCnt)
 	}
 
+	var sep string
+	if len(rule.Arguments) > 3 {
```
csuzhangxc: can we only do this check _once_ for one rule? like do it in `AddRule`/`UpdateRule `, not in `computePartitionID`.

pkg/column-mapping/column.go
```diff
@@ -63,10 +63,12 @@ var Exprs = map[Expr]func(*mappingInfo, []interface{}) ([]interface{}, error){
 	// # 3 table ID (table suffix)
 	// # 4 origin ID (>= 0, <= 17592186044415)
 	//
-	// others: schema = arguments[1] + schema suffix
-	//         table = arguments[2] + table suffix
-	//  example: schema = schema_1 table = t_1  => arguments[1] = "schema_", arguments[2] = "t_"
+	// others: schema = arguments[1]  or  arguments[1] + arguments[3] + schema suffix
```
IANTHEREAL, line 66: can you help to update readme? @kennytm

pkg/column-mapping/column.go
```diff
@@ -178,37 +182,33 @@ func (m *Mapping) AddRule(rule *Rule) error {
 	if !m.caseSensitive {
 		rule.ToLower()
 	}
+	if rule.Expression == PartitionID && len(rule.Arguments) == 3 {
```
IANTHEREAL: is it better to put it into `rule.Ajust`?

IANTHEREAL:
Rest LGTM

IANTHEREAL:
LGTM @csuzhangxc PTAL

csuzhangxc (approved):
LGTM. Could you help to update the docs-cn or docs?

kennytm:
pingcap/docs#1166.

PR #237, opened 2019-05-04
diff: add connection manage

WangXiangUSTC:
1.  a sql.DB may contains more than one connection, if execute set by sql.DB, will only set for one connection
2. when set snapshot, can't write data, so can't save checkpoint, should add another connection for it

1. add a new struct Conns to manage several connections, some connections used for compare data, and one connection for save checkpoint.
2. refine the log when find different data by the way

TODO: will add an integration test for snapshot later

### Check List 

Tests 

 - Unit test

Related changes


WangXiangUSTC:
@csuzhangxc @GregoryIan PTAL

csuzhangxc:
@WangXiangUSTC can we use [System Variables](https://github.com/go-sql-driver/mysql#system-variables) in DSN to set `tidb_snapshot` for a `sql.DB` object?

if so, we need two `sql.DB` objects, one for comparing data and another for saving the checkpoint.

WangXiangUSTC:
you are right, I add tidb_snapshot in DSN, and it works. PTAL again
and I remove sync log by the way, because these tools don't set log file, out put log to stdout, so don't need sync log. @csuzhangxc

pkg/diff/conn.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	// SetMaxOpenConns and SetMaxIdleConns for connection to avoid error like
+	// `dial tcp 10.26.2.1:3306: connect: cannot assign requested address`
+	db.SetMaxOpenConns(num)
+	db.SetMaxIdleConns(num)
+
+	cpDB, err := dbutil.OpenDB(dbConfig)
+	if err != nil {
+		return nil, errors.Errorf("create db connections %+v error %v", dbConfig, err)
```
csuzhangxc, line 45: if open `db` without `err`, but open `cpDB` with `err`, how should we close them?
close `db` in this `NewConns` or in `Close`? if in `Close`, maybe we need to check `cpDB != nil`.

WangXiangUSTC: add check https://github.com/pingcap/tidb-tools/pull/237/commits/5603aa2b7781c2a8da6c900b360a64349587860a

sync_diff_inspector/diff.go
```diff
@@ -95,41 +95,22 @@ func (df *Diff) init(cfg *Config) (err error) {
 	return nil
 }
 
+// CreateDBConn creates db connections for source and target.
 func (df *Diff) CreateDBConn(cfg *Config) (err error) {
-	// SetMaxOpenConns and SetMaxIdleConns for connection to avoid error like
-	// `dial tcp 10.26.2.1:3306: connect: cannot assign requested address`
 	for _, source := range cfg.SourceDBCfg {
-		source.Conn, err = dbutil.OpenDB(source.DBConfig)
+		source.Conns, err = diff.NewConns(df.ctx, source.DBConfig, cfg.CheckThreadCount, source.Snapshot)
```
csuzhangxc: if `err != nil`, we still needed to call `Close` in some place now?

pkg/diff/util.go
```diff
@@ -90,3 +93,18 @@ func rowContainsCols(row map[string]*dbutil.ColumnData, cols []*model.ColumnInfo
...
 }
+
+func rowToString(row map[string]*dbutil.ColumnData) string {
+	var s strings.Builder
+	s.WriteString("{ ")
+	for key, val := range row {
+		if val.IsNull {
+			s.WriteString(fmt.Sprintf("%s: IsNull, ", key))
+		} else {
+			s.WriteString(fmt.Sprintf("%s: %s, ", key, val.Data))
```
csuzhangxc, line 104: are you planning to convert it to a _JSON_ string? It not accurate, like the ending `,` and may some escaped `:`/`,`, but may not a problem for logging?

csuzhangxc (commented):
rest LGTM

pkg/diff/util.go
WangXiangUSTC: just for log, it doesn't matter

sync_diff_inspector/diff.go
WangXiangUSTC: already check DB is nil when close connections

csuzhangxc: if `diff.NewConns` `OpenDB` for `cpDB` failed, it will return `nil`, seems can not close `db` in it?

WangXiangUSTC: my mistake, fixed in https://github.com/pingcap/tidb-tools/pull/237/commits/c6e07faa140ca32548f9838719478ded193c65f7

csuzhangxc (commented):
~~LGTM~~

csuzhangxc:
@WangXiangUSTC you need to fix the CI, seems caused by the iter order of the map?

WangXiangUSTC:
test fixed, PTAL again @csuzhangxc

csuzhangxc (commented):
LGTM

csuzhangxc:
@GregoryIan PTAL

pkg/dbutil/common.go
```diff
@@ -110,7 +110,18 @@ func GetDBConfigFromEnv(schema string) DBConfig {
 // OpenDB opens a mysql connection FD
 func OpenDB(cfg DBConfig) (*sql.DB, error) {
 	dbDSN := fmt.Sprintf("%s:%s@tcp(%s:%d)/?charset=utf8mb4", cfg.User, cfg.Password, cfg.Host, cfg.Port)
-	dbConn, err := sql.Open("mysql", dbDSN)
+	return OpenDBByDSN(dbDSN)
+}
+
+// OpenDBWithSnapshot opens a mysql connection FD with tidb_snapshot set
+func OpenDBWithSnapshot(cfg DBConfig, snapshot string) (*sql.DB, error) {
```
IANTHEREAL: how about adding `Snapshot` into `DBConfig`?

IANTHEREAL: then we can save one function

pkg/diff/checkpoint_test.go
```diff
@@ -26,13 +27,18 @@ var _ = Suite(&testCheckpointSuite{})
...
+	defer conns.Close()
+
+	s.testInitAndGetSummary(c, conns.CpDB)
+	s.testSaveAndLoadChunk(c, conns.CpDB)
+	s.testUpdateSummary(c, conns.CpDB)
 
-	s.testInitAndGetSummary(c, db)
-	s.testSaveAndLoadChunk(c, db)
-	s.testUpdateSummary(c, db)
+	dropCheckpoint(ctx, conns.CpDB)
```
IANTHEREAL: it's better to call `defer dropCheckpoint(ctx, conns.CpDB)` at L35

WangXiangUSTC:
@GregoryIan PTAL again

pkg/diff/conn.go
```diff
@@ -0,0 +1,56 @@
...
+	// `dial tcp 10.26.2.1:3306: connect: cannot assign requested address`
+	db.SetMaxOpenConns(num)
+	db.SetMaxIdleConns(num)
+
+	return db, nil
+}
+
+// CreateDBForCP creates sql.DB used for write data for checkpoint
+func CreateDBForCP(ctx context.Context, dbConfig dbutil.DBConfig) (cpDB *sql.DB, err error) {
+	snapshot := dbConfig.Snapshot
```
GregoryIan: it's not safe, just clone one

IANTHEREAL: it's not a ponter, I think you don't need to restore snapshot field

pkg/diff/diff_test.go
```diff
@@ -244,11 +249,11 @@ func createTableDiff(db *sql.DB, sourceTableNames []string, targetTableName stri
 	}
 }
 
-func createConn() (*sql.DB, error) {
+func createConn(ctx context.Context) (*sql.DB, error) {
```
IANTHEREAL: remove `ctx context.Context`?

WangXiangUSTC: done

pkg/diff/conn.go
WangXiangUSTC: done

IANTHEREAL:
any test except for unit test？ @WangXiangUSTC

IANTHEREAL (approved):
LGTM

WangXiangUSTC:
will add a integration test next pr @GregoryIan

PR #240, opened 2019-05-24
diff: remove use-rowid config

WangXiangUSTC:
row_id is not suggest used in drainer, so don't need this config in diff

remove use-rowid config, and some related function

### Check List 

Tests 

 - Unit test

Related changes


csuzhangxc (commented):
LGTM

IANTHEREAL (approved):
LGTM

Issue #234, opened 2019-04-23
remove-columns cause index lookup error


amyangfei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?

run sync_diff in sharding mode, remove-columns option changes the columns in table info slice, while index column keeps the table offset, which leads to index out of range error

```
[[check-tables]]
schema = "db_target"
tables = ["t_target"]

[[table-config]]
schema = "db_target"
table = "t_target"
remove-columns = ["id"]
is-sharding = true
index-fields = "uid"

[[table-config.source-tables]]
instance-id = "source-1"
schema = "sharding1"
table  = "~t.*"

[[table-config.source-tables]]
instance-id = "source-2"
schema = "sharding1"
table  = "~t.*"
```

```
mysql> show create table t1\G
*************************** 1. row ***************************
       Table: t1
Create Table: CREATE TABLE `t1` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `uid` int(11) DEFAULT NULL,
  `name` varchar(80) DEFAULT NULL,
  `info` varchar(100) DEFAULT NULL,
  `age` int(11) DEFAULT NULL,
  `info_json` json GENERATED ALWAYS AS (`info`) VIRTUAL,
  `id_gen` int(11) GENERATED ALWAYS AS ((`uid` + 1)) VIRTUAL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `uid` (`uid`),
  UNIQUE KEY `id_gen` (`id_gen`),
  KEY `multi_col_idx` (`uid`,`id_gen`)
) ENGINE=InnoDB AUTO_INCREMENT=4 DEFAULT CHARSET=utf8mb4
1 row in set (0.00 sec)
```

2. What did you expect to see?

diff runs successfully

3. What did you see instead?

```
panic: runtime error: index out of range

goroutine 42 [running]:
github.com/pingcap/tidb-tools/pkg/dbutil.SelectUniqueOrderKey(0xc0003ba500, 0x0, 0x20, 0x203000, 0x203000, 0x0, 0x0)
        /home/apple/work/code/pingcap/gopath/src/github.com/pingcap/tidb-tools/pkg/dbutil/index.go:183 +0xb3b
[...]

PR #242, opened 2019-05-27
diff: some minor fix

WangXiangUSTC:
1. fix issue https://github.com/pingcap/tidb-tools/issues/234
2. when fix sql contain generated column can't execute in mysql/tidb
3. when column's value have `'`, the fix sql don't have escape charset, so can't execute in mysql/tidb

fix these bugs

### Check List 

Tests 

 - Unit test
 
Related changes

### DIFF
--- a/pkg/dbutil/table.go
+++ b/pkg/dbutil/table.go
@@ -134,6 +134,8 @@ func columnDefToCol(offset int, colDef *ast.ColumnDef) (*model.ColumnInfo, []*as
 			case ast.ColumnOptionComment:
 				// do nothing
 			case ast.ColumnOptionGenerated:
+				// FIXME: use a default string now to make col.IsGenerated() return true, will use real generated expr string later
+				col.GeneratedExprString = "Generated"
 				// do nothing
 			case ast.ColumnOptionFulltext:
 				// do nothing
--- a/pkg/diff/diff_test.go
+++ b/pkg/diff/diff_test.go
@@ -35,7 +35,7 @@ var _ = Suite(&testDiffSuite{})
 type testDiffSuite struct{}
 
 func (*testDiffSuite) TestGenerateSQLs(c *C) {
-	createTableSQL := "CREATE TABLE `test`.`atest` (`id` int(24), `name` varchar(24), `birthday` datetime, `update_time` time, `money` decimal(20,2), primary key(`id`))"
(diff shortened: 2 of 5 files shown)


pkg/diff/diff_test.go
```diff
@@ -75,6 +76,11 @@ func (*testDiffSuite) TestGenerateSQLs(c *C) {
 	deleteSQL = generateDML("delete", rowsData, orderKeyCols, tableInfo, "test")
 	c.Assert(replaceSQL, Equals, "REPLACE INTO `test`.`atest`(`id`,`name`,`birthday`,`update_time`,`money`) VALUES (NULL,NULL,'2018-01-01 00:00:00','10:10:10',11.1111);")
 	c.Assert(deleteSQL, Equals, "DELETE FROM `test`.`atest` WHERE `id` is NULL;")
+
+	// test value with "'"
+	rowsData["name"] = &dbutil.ColumnData{Data: []byte("a'a"), IsNull: false}
+	replaceSQL = generateDML("replace", rowsData, orderKeyCols, tableInfo, "test")
```
csuzhangxc, line 82: is it needed to test `generateDML("delete" ...`

WangXiangUSTC: done

csuzhangxc (commented):
LGTM

pkg/diff/util.go
```diff
@@ -64,6 +66,24 @@ func removeColumns(tableInfo *model.TableInfo, columns []string) *model.TableInf
 		}
 	}
 
+	// calculate column offset
+	colMap := make(map[string]int)
```
suzaku:
```suggestion
	colMap := make(map[string]int, len(tableInfo.Columns))
```

pkg/dbutil/table.go
```diff
@@ -134,6 +134,8 @@ func columnDefToCol(offset int, colDef *ast.ColumnDef) (*model.ColumnInfo, []*as
 			case ast.ColumnOptionComment:
 				// do nothing
 			case ast.ColumnOptionGenerated:
+				// FIXME: use a default string now to make col.IsGenerated() return true, will use real generated expr string later
+				col.GeneratedExprString = "Generated"
```
suzaku, line 138: What do you need to do to get the real value for this field?

WangXiangUSTC: don't need it now,  can judge this column is generated is ok, maybe we need the real value later

suzaku (approved):
LGTM

PR #245, opened 2019-06-05
diff: update column's length for checkpoint table

WangXiangUSTC:
user meet error `Check data difference failed "Error 1406: Data too long, field len 30, data len 31"`
need increase some columns' length in checkpoint table

update some columns' length from 30 to 64（the max schema/table name length in mysql and tidb https://dev.mysql.com/doc/refman/8.0/en/identifiers.html）


suzaku (commented):
LGTM

amyangfei (approved):
LGTM

PR #248, opened 2019-06-12
diff: update ignore-column config && add integration test

WangXiangUSTC:
there is something wrong with config `remove-column` and `ignore-column`

1. delete config `remove-column`, and update `ignore-column`
2. when ignore a column in one index, ignore this index
3. add integration test

### Check List 

Tests 

 - Integration test


pkg/diff/README.md
```diff
@@ -16,14 +16,9 @@ type TableDiff struct {
...
 	TargetTable *TableInstance
 
-	// columns be ignored, will not check this column's data, 
-	// but may use these columns as split field or order by key.
+	// columns be ignored, will not check this column's data
 	IgnoreColumns []string
 
-	// columns be removed, will remove these columns from table info,
-	// and will not check these columns' data, will not use these columns as split field or order by key too.
-	RemoveColumns []string
```
WangXiangUSTC, line 25: only when column's value is different in TiDB and MySQL need set ignore-column. If data is different and use this column to split chunk, the chunk's range in TiDB and MySQL is also different. so we don't need both have ignore-column and remove-column, I delete the remove-column, and use ignore-column to replace remove-column.

pkg/diff/util.go
```diff
@@ -48,12 +48,9 @@ func removeColumns(tableInfo *model.TableInfo, columns []string) *model.TableInf
 		for j := 0; j < len(index.Columns); j++ {
 			col := index.Columns[j]
 			if _, ok := removeColMap[col.Name.O]; ok {
-				index.Columns = append(index.Columns[:j], index.Columns[j+1:]...)
-				j--
-				if len(index.Columns) == 0 {
-					tableInfo.Indices = append(tableInfo.Indices[:i], tableInfo.Indices[i+1:]...)
-					i--
-				}
+				tableInfo.Indices = append(tableInfo.Indices[:i], tableInfo.Indices[i+1:]...)
```
WangXiangUSTC, line 51: when remove a column in an index, also remove this index.

pkg/diff/diff.go
```diff
@@ -690,7 +686,7 @@ func generateDML(tp string, data map[string]*dbutil.ColumnData, keys []*model.Co
 		sql = fmt.Sprintf("REPLACE INTO `%s`.`%s`(%s) VALUES (%s);", schema, table.Name, strings.Join(colNames, ","), strings.Join(values, ","))
 	case "delete":
 		kvs := make([]string, 0, len(keys))
-		for _, col := range keys {
+		for _, col := range table.Columns {
```
WangXiangUSTC, line 689: for example:
have two chunk: chunk1's range is a = 1 and b <= 1, chunk2's range is a = 1 and b > 1, and a is the primary key, b is column with index.
in source have one row a=1, b = 0, and in target have a wrong row  a = 1, b =2
when compare chunk1 ,will generate a fix sql sql1: replace into test.t1 values(1, 0),
when compare chunk2, will generate a fix sql sql2 if where only contain pk/uk: delete test.t1 where a = 1

when user execute these fix sqls, and sql1 is before sql2 will delete this row.
so need contain all the column's value in delete sql.

IANTHEREAL:
please refine description

pkg/diff/diff.go
```diff
@@ -188,14 +185,14 @@ func (t *TableDiff) getTableInfo(ctx context.Context) error {
 	if err != nil {
 		return errors.Trace(err)
 	}
-	t.TargetTable.info = removeColumns(tableInfo, t.RemoveColumns)
+	t.TargetTable.info = removeColumns(tableInfo, t.IgnoreColumns)
```
IANTHEREAL: if this column is index column, I think we should remove this index all in some case, like remove primary column of this index. how do you think about it?

IANTHEREAL:
Rest LGTM

IANTHEREAL:
please resolve conflicts and refine the description

pkg/diff/diff.go
WangXiangUSTC: we use tidb's statistical information which is treat index as a unit to split chunk, so if removed any column in an index, this index's  statistical information is unavailable.

WangXiangUSTC:
@GregoryIan resolved, PTAL again

IANTHEREAL:
LGTM

pkg/diff/diff.go
csuzhangxc: @WangXiangUSTC so, can we remove the `keys` arg now?

pkg/diff/util_test.go
```diff
@@ -23,7 +23,7 @@ var _ = Suite(&testUtilSuite{})
 
 type testUtilSuite struct{}
 
-func (s *testUtilSuite) TestRemoveColumns(c *C) {
+func (s *testUtilSuite) TestIgnoreColumns(c *C) {
```
csuzhangxc, line 26: do we need to rename `removeColumns` to `ignoreColumns` in _util.go_?

sync_diff_inspector/diff.go
```diff
@@ -211,7 +212,6 @@ func (df *Diff) AdjustTableConfig(cfg *Config) (err error) {
 					Schema: schemaTables.Schema,
 					Table:  tableName,
 				},
-				IgnoreColumns:   make([]string, 0, 1),
```
csuzhangxc, line 214: do not init it anymore? init directly in L261?

tests/sync_diff_inspector/ignore_column/run.sh
```diff
@@ -0,0 +1,27 @@
...
+
+echo "update data in column b, data should not be equal"
+mysql -uroot -h 127.0.0.1 -P 4000 -e "update diff_test.test set b = 'abc' limit 1"
+
+rm $OUT_DIR/fix.sql || true
+sync_diff_inspector --config=./config_base.toml > $OUT_DIR/ignore_column_diff.log || true
+check_contains "sourceDB don't equal targetDB" $OUT_DIR/ignore_column_diff.log
+mv $OUT_DIR/fix.sql $OUT_DIR/fix.sql.bak
+
+echo "ignore check column c, check result should be pass"
```
csuzhangxc: `c` or `b`?

csuzhangxc (commented):
rest LGTM

pkg/diff/diff.go
WangXiangUSTC: removed

sync_diff_inspector/diff.go
WangXiangUSTC: yes, in fact golang will initial it

tests/sync_diff_inspector/ignore_column/run.sh
WangXiangUSTC: updated

pkg/diff/util_test.go
WangXiangUSTC: name all changed

csuzhangxc (approved):
LGTM

PR #249, opened 2019-06-13
diff: remove tidb-instance-id config and find tidb instance automatic

WangXiangUSTC:
can find tidb instance by select version(), so don't need user to set tidb-instance-id manually.

1. remove tidb-instance-id config and find tidb instance by IsTiDB function
2. execute analyze table in integration test, so can use tidb's statistics information to split chunk

### Check List 

Tests 
 - Integration test


IANTHEREAL:
LGTM

csuzhangxc (approved):
LGTM

PR #252, opened 2019-06-21
pkg: use case insensitive event in binlog filter

amyangfei:

binlog event is lowercase only now. It is not harmful to be case insensitive.

change the input rule to lowercase EventType before match.

### Check List 

Tests 

 - Unit test

### DIFF
--- a/pkg/binlog-filter/filter.go
+++ b/pkg/binlog-filter/filter.go
@@ -265,7 +265,9 @@ func (b *BinlogEvent) Filter(schema, table string, event EventType, rawQuery str
 }
 
 func (b *BinlogEvent) matchEvent(tp, event EventType, rules []EventType) bool {
-	for _, rule := range rules {
+	for _, rawRule := range rules {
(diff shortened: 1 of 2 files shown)


amyangfei:
PTAL @GregoryIan @csuzhangxc

pkg/binlog-filter/filter.go
```diff
@@ -265,7 +265,9 @@ func (b *BinlogEvent) Filter(schema, table string, event EventType, rawQuery str
 }
 
 func (b *BinlogEvent) matchEvent(tp, event EventType, rules []EventType) bool {
-	for _, rule := range rules {
+	for _, rawRule := range rules {
+		rule := EventType(strings.ToLower(string(rawRule)))
```
csuzhangxc: can we do this transformation when adding/updating rules?

csuzhangxc (commented):
LGTM

WangXiangUSTC (approved):
LGTM

PR #254, opened 2019-07-09
Fix: watcher's modify event condition won't trigger on macOS

3pointer:
watcher on macOS (HFS+) won't send modify event in some case due to the file may change in the same second and HFS+ only stores [second level timestamp](https://stackoverflow.com/questions/18403588/how-to-return-millisecond-information-for-file-access-on-mac-os-x-in-java)

```package main
import "fmt"
import "os"

func main() {
    fileName := "test"
    f, _ := os.OpenFile(fileName, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
    stat, _ := os.Stat(fileName)
    for i := 0; i < 10; i ++ {
        f.Write([]byte("test\n"))
    }
    stat2, _ := os.Stat(fileName)
    fmt.Println(stat.ModTime() == stat2.ModTime())
}
```

code above may return true on my computer sometimes
Use ModTime() + Size() can reduce error scenarios on macOS

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/pkg/watcher/watcher.go
+++ b/pkg/watcher/watcher.go
@@ -198,7 +198,10 @@ func (w *Watcher) pollEvents(currFileList map[string]os.FileInfo) {
 			continue
 		}
 
-		if !latestFi.ModTime().Equal(currFi.ModTime()) {
+		// ModTime may return timestamp in second level on some file system
+		// So use ModTime + Size to judge modify event will be more precisely
+		if !latestFi.ModTime().Equal(currFi.ModTime()) ||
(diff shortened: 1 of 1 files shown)


csuzhangxc (commented):
LGTM. 
@amyangfei PTAL

amyangfei (approved):
LGTM

PR #255, opened 2019-07-11
refine reader.go to avoid message blocking

lichunzhu:
Refine kafka reader. The current version may block when the consumer thread exits.
Issue link: [TOOL-1402](https://internal.pingcap.net/jira/browse/TOOL-1402)

Add one more select to listen signal from both downstream and stop channel.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/tidb-binlog/driver/reader/reader.go
+++ b/tidb-binlog/driver/reader/reader.go
@@ -154,13 +154,15 @@ func (r *Reader) run() {
 
 	// use variable msg to avoid message blocking while reading
 	var msg *Message
+	var destChan chan *Message
+	var sourceChan <-chan *sarama.ConsumerMessage
 	for {
-		var msgM chan *Message
-		var msgC <-chan *sarama.ConsumerMessage
+		destChan = nil
+		sourceChan = nil
 		if msg != nil {
-			msgM = r.msgs
+			destChan = r.msgs
 		} else {
-			msgC = partitionConsumer.Messages()
+			sourceChan = partitionConsumer.Messages()
 		}
 		select {
 		case <-r.stop:
@@ -168,7 +170,7 @@ func (r *Reader) run() {
 			close(r.msgs)
 			log.Info("reader stop to run")
 			return
-		case kmsg := <-msgC:
+		case kmsg := <-sourceChan:
 			log.Debug("get kafka message", zap.Int64("offset", kmsg.Offset))
 			binlog := new(pb.Binlog)
 			err := binlog.Unmarshal(kmsg.Value)
@@ -185,7 +187,7 @@ func (r *Reader) run() {
 				Binlog: binlog,
 				Offset: kmsg.Offset,
 			}
-		case msgM <- msg:
+		case destChan <- msg:
 			msg = nil
 		}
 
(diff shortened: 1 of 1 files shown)


lichunzhu:
@july2993 PTAL

tidb-binlog/driver/reader/reader.go
```diff
@@ -152,14 +152,23 @@ func (r *Reader) run() {
 	}
 	defer partitionConsumer.Close()
 
+	// use variable msg to avoid message blocking while reading
+	var msg *Message
 	for {
+		var msgM chan *Message
+		var msgC <-chan *sarama.ConsumerMessage
+		if msg != nil {
+			msgM = r.msgs
```
july2993: once reach this code, msgM is always r.msgs

tidb-binlog/driver/reader/reader.go
```diff
@@ -152,14 +152,23 @@ func (r *Reader) run() {
...
+		} else {
+			msgC = partitionConsumer.Messages()
```
july2993: once reach this code msgC is alwasy `partitionConsumer.Messages()`

july2993 (commented):
pls use the more descriptive variable name

tidb-binlog/driver/reader/reader.go
lichunzhu: revised

tidb-binlog/driver/reader/reader.go
lichunzhu: /home/pingcap/workspace2/tidb-binlog

lichunzhu: revised

tidb-binlog/driver/reader/reader.go
```diff
@@ -172,10 +183,12 @@ func (r *Reader) run() {
 				continue
 			}
 
-			r.msgs <- &Message{
+			msg = &Message{
 				Binlog: binlog,
 				Offset: kmsg.Offset,
 			}
```
suzaku, line 180: The problem here is that, after the caller of `reader` stops reading from `r.msgs`, this channel may become full and block this line and the whole `run` goroutine even after closed.

One way to avoid this is to add another `select` here with `<-r.stop` as a case and `continue` if `r.stop` is closed.

july2993: the current solution also works, right?

suzaku:
```diff
diff --git a/tidb-binlog/driver/reader/reader.go b/tidb-binlog/driver/reader/reader.go
index 726f064..950545c 100644
--- a/tidb-binlog/driver/reader/reader.go
+++ b/tidb-binlog/driver/reader/reader.go
@@ -172,10 +172,16 @@ func (r *Reader) run() {
 				continue
 			}
 
-			r.msgs <- &Message{
+			msg := &Message{
 				Binlog: binlog,
 				Offset: kmsg.Offset,
 			}
+			select {
+			case r.msgs <- msg:
+			case <-r.stop:
+				// In the next iteration, the <-r.stop would match again and prepare to quit
+				continue
+			}
 		}
 
 	}
```

@lichunzhu PTAL

suzaku:
LGTM

There seems to be something wrong with the unit testing environment.

IANTHEREAL (approved):
LGTM

PR #256, opened 2019-07-16
diff: split bucket by random if the bucket's count is twice more than chunk-size

WangXiangUSTC:
# What problem does this PR solve? 
fix issue: https://internal.pingcap.net/jira/browse/TOOL-1372
tidb has a max value for bucket's num, so if one table have too many rows, one bucket will contains many rows, and may cause oom when select data.

1. refine the random spliter, use same range expression as bucket spliter
2. when a bucket's count is twice bigger than chunk-size, use random split function to split it

### Check List 

Tests 

 - Unit test


pkg/dbutil/common.go
```diff
@@ -190,20 +190,18 @@ func GetRowCount(ctx context.Context, db *sql.DB, schemaName string, tableName s
...
-		+------+-------+
-		| id   | count |
-		+------+-------+
-		|    1 |     2 |
-		|    2 |     2 |
-		|    3 |     1 |
-		+------+-------+
-
-		FIXME: TiDB now don't return rand value when use `ORDER BY RAND()`
+		mysql> SELECT `id` FROM (SELECT `id`, rand() rand_value FROM `test`.`test`  WHERE `id` COLLATE "latin1_bin" > 0 AND `id` COLLATE "latin1_bin" < 100 ORDER BY rand_value LIMIT 5) rand_tmp ORDER BY `id` COLLATE "latin1_bin";
```
WangXiangUSTC, line 200: old sql will not select random value in tidb, update it to suggest sql, https://github.com/pingcap/tidb/issues/9033

IANTHEREAL: any test for it?  this kind of mistake is too bad.

pkg/diff/chunk.go
```diff
@@ -85,32 +81,10 @@ func (c *ChunkRange) toString(collation string) (string, []string) {
...
-		}
-
-		return strings.Join(conditions, " AND "), args
-	}
-
 	/* for example:
 	there is a bucket in TiDB, and the lowerbound and upperbound are (v1, v3), (v2, v4), and the columns are `a` and `b`,
-	this bucket's data range is (a > v1 or (a == v1 and b >= v2)) and (a < v3 or (a == v3 and a <= v4)),
-	not (a >= v1 and a <= v3 and b >= v2 and b <= v4)
+	this bucket's data range is (a > v1 or (a == v1 and b > v3)) and (a < v2 or (a == v2 and a <= v4)),
```
IANTHEREAL: why change it?

pkg/dbutil/common.go
WangXiangUSTC: you means add test in tidb or in diff?

pkg/diff/chunk.go
WangXiangUSTC: the old comment is wrong, v1 and v2 is column a's range, v3 and v4 is column b's range. And I change the chunk's range's expression, so need change this example.

pkg/diff/chunk.go
```diff
@@ -123,25 +97,31 @@ func (c *ChunkRange) toString(collation string) (string, []string) {
 	preConditionArgsForLower := make([]string, 0, 1)
 	preConditionArgsForUpper := make([]string, 0, 1)
 
-	for _, bound := range c.Bounds {
-		if len(bound.Lower) != 0 {
+	for i, bound := range c.Bounds {
+		lowerSymbol := gt
+		upperSymbol := lt
+		if i == len(c.Bounds)-1 {
```
GregoryIan:
```suggestion
		if i < len(c.Bounds)-1 {
```

pkg/diff/chunk.go
IANTHEREAL:
```suggestion
	this bucket's data range is (a > v1 or (a == v1 and b > v3)) and (a < v2 or (a == v2 and b <= v4)),
```

pkg/diff/chunk.go
```diff
@@ -221,170 +205,57 @@ func (s *randomSpliter) split(table *TableInstance, columns []*model.ColumnInfo,
...
 
 	return chunks, nil
 }
 
-// splitRange splits a chunk to multiple chunks.
-func (s *randomSpliter) splitRange(db *sql.DB, chunk *ChunkRange, count int, schema string, table string, columns []*model.ColumnInfo) ([]*ChunkRange, error) {
-	var chunks []*ChunkRange
-
+// splitRangeByRandom splits a chunk to multiple chunks by random
+func splitRangeByRandom(db *sql.DB, chunk *ChunkRange, count int, schema string, table string, columns []*model.ColumnInfo, limits, collation string) (chunks []*ChunkRange, err error) {
```
IANTHEREAL, line 244: I hope you can test the efficiency of splitting under various cases.

IANTHEREAL: because I don't know whether it's better

pkg/dbutil/common.go
IANTHEREAL: It would make `diff` tools low availability,  so how can we quickly find this problem next time? I understand that there is a lack of performance assessment.

pkg/diff/chunk.go
IANTHEREAL: is the old implementation wrong

pkg/diff/chunk.go
WangXiangUSTC: ok, I will do a performance test

pkg/diff/chunk.go
WangXiangUSTC: no, just comment is wrong.

WangXiangUSTC: fixed

pkg/check/table_structure.go
```diff
@@ -20,14 +20,13 @@ import (
...
 
-	"github.com/pingcap/tidb-tools/pkg/column-mapping"
-
 	"github.com/pingcap/errors"
 	"github.com/pingcap/parser"
 	"github.com/pingcap/parser/ast"
 	"github.com/pingcap/parser/charset"
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/parser/mysql"
+	column "github.com/pingcap/tidb-tools/pkg/column-mapping"
```
csuzhangxc, line 29: redundant alias (`column`)?

pkg/diff/chunk.go
```diff
@@ -162,43 +141,48 @@ func (c *ChunkRange) toString(collation string) (string, []string) {
...
-		UpperSymbol: upperSymbol,
-	}
-
+func (c *ChunkRange) update(column, lower, upper string) {
 	for i, b := range c.Bounds {
 		if b.Column == column {
 			// update the bound
-			c.Bounds[i] = newBound
+			if len(lower) > 0 {
+				c.Bounds[i].Lower = lower
```
csuzhangxc: do not support to overwrite it with an empty value (`lower`/`upper`)?
if you just do not want to overwrite the original value, how about pass the original as the argument?

pkg/diff/chunk.go
```diff
@@ -221,170 +205,57 @@ func (s *randomSpliter) split(table *TableInstance, columns []*model.ColumnInfo,
...
-	lowerSymbol := symbolMin
-	upperSymbol := lt
-
-	for i := 0; i < len(splitValues); i++ {
-		if i == 0 && useNewColumn {
-			// create chunk less than min
-			newChunk := chunk.copyAndUpdate(splitCol, "", "", splitValues[i], lt)
-			chunks = append(chunks, newChunk)
-		}
+	for i := 0; i <= len(randomValues[0]); i++ {
```
csuzhangxc: Can we always get `num` values for all columns? In other words, all slices in `randomValues` always have the same length? (same question for L315)

BTW, we need a 4-level `for` loop (with two in `ChunkRange.update`) to construct these `chunks`, can simply it (construct without `update`)?

pkg/diff/chunk.go
```diff
@@ -423,43 +292,63 @@ func (s *bucketSpliter) getChunksByBuckets() ([]*ChunkRange, error) {
...
+					if len(upperValues) > 0 {
+						upperValue = upperValues[j]
+					}
+					chunk.update(column.Name.O, lowerValue, upperValue)
 				}
+			}
+
+			if i == len(buckets) {
+				chunks = append(chunks, chunk)
+				continue
```
csuzhangxc: can we `break` here?

pkg/diff/chunk.go
```diff
@@ -423,43 +292,63 @@ func (s *bucketSpliter) getChunksByBuckets() ([]*ChunkRange, error) {
...
+			}
 
+			count := (buckets[i].Count - latestCount) / int64(s.chunkSize)
+			if count == 0 {
+				continue
```
csuzhangxc, line 360: if `buckets[i].Count - latestCount == 99` and `s.chunkSize == 100`, then `count == 0`, is a chunk need to be append for this?

pkg/diff/chunk.go
```diff
@@ -502,19 +391,22 @@ func getSplitFields(table *model.TableInfo, splitFields []string) ([]*model.Colu
...
+	}
 
-	// user's config had higher priorities
-	for _, col := range append(append(splitCols, indexColumns...), table.Columns...) {
-		if _, ok := colsMap[col.Name.O]; ok {
-			continue
+	for _, col := range table.Columns {
+		colsMap[col.Name.O] = col
+	}
+	indices := dbutil.FindAllIndex(table)
```
csuzhangxc, line 422: Do indices (returned from `FindAllIndex`) including some indices with some **ignored** columns?

pkg/check/table_structure.go
WangXiangUSTC: no, if execute gofmt will remove this line because this package name is `column` not `column-mapping`

csuzhangxc: :open_mouth: I tried, but it seems not removed.

WangXiangUSTC: I don't know why. And another reason, the package name is different with the directory name, if some one what to update this code, may don't know where this pkg is used.

pkg/diff/chunk.go
WangXiangUSTC: can break, updated

pkg/diff/chunk.go
WangXiangUSTC: may be don't have same num values, because get random value for every column one by one, and the data may be changed. but we define a limit for diff: don't support online check. And I still add a new function `minLenInSlices` to get the min length in `[][]string` in 5bcf5bc95eb4091e977f2fa15294eb56035c43b6

WangXiangUSTC: for L315, don't have this question

pkg/diff/util_test.go
```diff
@@ -96,3 +96,32 @@ func (s *testUtilSuite) TestRowToString(c *C) {
...
+			[][]string{
+				{"1", "2"},
+				{"1", "2", "3"},
+			},
+			2,
+		}, {
+			[][]string{
+				{"1", "2"},
+				{},
+			},
```
csuzhangxc, line 113: what is the min length of 
```go
[][]string {
	{},
	{"1", "2'},
}
```

WangXiangUSTC: a mistake, fixed in 5389e91a1bb81777301070c681ff9c1a1d453933

pkg/diff/chunk.go
WangXiangUSTC: no, the table info already removed column and index with this column

pkg/diff/chunk.go
WangXiangUSTC: https://github.com/pingcap/tidb-tools/pull/256/commits/0ad3dcfa1ddf4073164486b5fe0ef0941a0ae94e
and a `columOffset` map in `ChunkRange` to remove a for loop.(only have one loop in `update` before)

pkg/diff/chunk.go
WangXiangUSTC: in fact it can't distinguish the lower/upper is "" or don't have lower/upper, I add `HasLower` and `HasUpper` in `Bound` in  https://github.com/pingcap/tidb-tools/pull/256/commits/040a4a27a08bb9494dba9c2924775a645b9bedc6

pkg/diff/chunk.go
WangXiangUSTC: I think don't need, the chunk size is only a reference value, just make the rule simple

IANTHEREAL:
LGTM

pkg/diff/chunk.go
WangXiangUSTC: I do a test, table with one hundred million rows, had 204 buckets in tidb's statistical information.
set the chunk-size to 1000,  diff get random value 209 times, and total cost 1m50s to split chunk.

csuzhangxc (approved):
LGTM

PR #258, opened 2019-07-24
diff: clean checkpoint info in several times to avoid `transaction is too large` error in TiDB

WangXiangUSTC:
fix issue: https://internal.pingcap.net/jira/browse/TOOL-1432
tidb only support delete less than 300,000 line one transaction,

1. delete 100,000 rows once
2. add check for chunk-size
3. update summary after checkpoint is loaded to avoid warn log

### Check List 

Tests 

 - Unit test


pkg/dbutil/common.go
```diff
@@ -756,3 +756,22 @@ func ignoreDDLError(err error) bool {
 		return false
 	}
 }
+
+// DeleteRows delete rows in several times. Only can delete less than 300,000 one time in TiDB.
+func DeleteRows(ctx context.Context, db *sql.DB, schemaName string, tableName string, where string, args []interface{}) error {
```
july2993, line 761: why not just truncate table

kennytm: TRUNCATE doesn't accept a `where` clause

pkg/dbutil/common_test.go
```diff
@@ -129,3 +131,19 @@ func (s *testDBSuite) TestIsIgnoreError(c *C) {
...
+	// delete and select twice
+	mock.ExpectExec("DELETE FROM").WillReturnResult(sqlmock.NewResult(0, 1))
+	countRows := sqlmock.NewRows([]string{"cnt"}).AddRow(199999)
+	mock.ExpectQuery("SELECT COUNT.*").WillReturnRows(countRows)
+	mock.ExpectExec("DELETE FROM").WillReturnResult(sqlmock.NewResult(0, 1))
+	countRows = sqlmock.NewRows([]string{"cnt"}).AddRow(0)
+	mock.ExpectQuery("SELECT COUNT.*").WillReturnRows(countRows)
+
+	err = DeleteRows(context.Background(), db, "test", "t", "", nil)
+	c.Assert(err, IsNil)
```
3pointer, line 144:
```
if err := mock.ExpectationsWereMet(); err != nil {
		c.Errorf("there were unfulfilled expectations: %s", err)
}
```

pkg/dbutil/common.go
```diff
@@ -756,3 +756,22 @@ func ignoreDDLError(err error) bool {
...
+	deleteSQL := fmt.Sprintf("DELETE FROM `%s`.`%s` WHERE %s limit 100000;", schemaName, tableName, where)
+	_, err := db.ExecContext(ctx, deleteSQL, args...)
```
july2993: can we just check RowsAffected instead of `GetRowCount`

WangXiangUSTC: https://github.com/pingcap/tidb-tools/pull/258/commits/bee3bb51f92908f70d171c6520124cdb5d2ab4db

pkg/dbutil/common.go
```diff
@@ -45,6 +45,9 @@ const (
 
 	// SlowWarnLog defines the duration to log warn log of sql when exec time greater than
 	SlowWarnLog = 100 * time.Millisecond
+
+	// DefaultDeleteRowsNum is the default rows num for delete one time
+	DefaultDeleteRowsNum = 100000
```
IANTHEREAL: is it reasonable?

pkg/dbutil/common.go
```diff
@@ -161,7 +164,7 @@ func GetCreateTableSQL(ctx context.Context, db *sql.DB, schemaName string, table
 
 // GetRowCount returns row count of the table.
 // if not specify where condition, return total row count of the table.
-func GetRowCount(ctx context.Context, db *sql.DB, schemaName string, tableName string, where string) (int64, error) {
+func GetRowCount(ctx context.Context, db *sql.DB, schemaName string, tableName string, where string, args []interface{}) (int64, error) {
```
IANTHEREAL: where to use `args`?

pkg/dbutil/common.go
IANTHEREAL:
```suggestion
	DefaultDeleteRowsNum int64 = 100000
```

pkg/diff/diff.go
```diff
@@ -168,7 +171,12 @@ func (t *TableDiff) CheckTableStruct(ctx context.Context) (bool, error) {
 
 func (t *TableDiff) adjustConfig() {
 	if t.ChunkSize <= 0 {
-		t.ChunkSize = 100
+		log.Warn("chunk size is less than 0, will use default value 1000", zap.Int("chunk size", t.ChunkSize))
+		t.ChunkSize = 1000
+	}
+
+	if t.ChunkSize < 1000 || t.ChunkSize > 10000 {
+		log.Warn("chunk size is suggest in range [1000, 10000]", zap.Int("chunk size", t.ChunkSize))
```
IANTHEREAL:
```suggestion
		log.Warn("chunk size is recommend in range [1000, 10000]", zap.Int("chunk size", t.ChunkSize))
```

IANTHEREAL:
Rest LGTM

pkg/dbutil/common.go
WangXiangUSTC: which number do you suggest?

pkg/dbutil/common.go
WangXiangUSTC: removed

IANTHEREAL:
LGTM

july2993 (approved):
LGTM

PR #259, opened 2019-08-07
diff: update column's type in checkpoint table

WangXiangUSTC:
the `range` field is a VARCHAR(100) which may be too short in some scenarios.

changing it to a TEXT field


kennytm (commented):
LGTM

amyangfei (approved):
LGTM

PR #244, opened 2019-06-04
pump_client: bug fix and enhancement...

july2993:
### What problem does this PR solve? 

We often encounter replication delay problems caused by the misacth of p-binlog and c-binlog in one pump  in internal tests.

there are two reasons
* c-binlog write failed. If the grpc connection is unavailable in the previous  implementation, then c-binlog will definitely fail to write. This is a bug.

* write more than one p-binlog in more than one pumps, but only write down one c-binlog, it would caused by pump client selector strategy

### What is changed and how it works?

1. don't recreate ClientConn unless Addr is change, ClientConn will maintain the connections to server and reconnect if need.
2,. only choose pump once, try all pumps at last

### Check List 

Tests 

 - Unit test
 - Integration test
Code changes

PR #260, opened 2019-08-08
pump_client: bugfixes

july2993:
drop the older messy pr https://github.com/pingcap/tidb-tools/pull/244
this pr only bugfix some explicit issue and don't make any logically changes, should be easy to review, may fire another pr after this.

1, Fix didn't set the choosePump at backoffWriteBinlog case in
WriteBinlog, this may make client don't write the c-binlog and pump lost
the c-binlog.
2, In backoffWriteBinlog, try every online state pump but the defined
`!pump.IsUsable`.
3, in watchStatus(), if the return err is ErrCompacted and you specify
the compacted revision again, will absolute return ErrCompacted again. This will make client can't no longer know new pumps online or pump status change.
4, in `detect()`,  detect every online pump but not `IsUsable()`
note before this pr, if `ErrNum` reach `defaultMaxErrNums`(10) it will not dectct the pump anymore(will add one to ErrNum once write detect fail, and the pump will be always unavailable unless the state in pd change(like stop pump and start again to make if change state)
```
// IsUsable returns true if pump is usable.
func (p *PumpStatus) IsUsable() bool {
    if !p.ShouldBeUsable() {
        return false
    }

    if atomic.LoadInt64(&p.ErrNum) > defaultMaxErrNums {
        return false
    }

    return true
}
```

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/tidb-binlog/pump_client/client.go
+++ b/tidb-binlog/pump_client/client.go
@@ -46,7 +46,7 @@ const (
 	// DefaultBinlogWriteTimeout is the default max time binlog can use to write to pump.
 	DefaultBinlogWriteTimeout = 15 * time.Second
 
-	// CheckInterval is the default interval for check unavaliable pumps.
+	// CheckInterval is the default interval for check unavailable pumps.
 	CheckInterval = 30 * time.Second
 )
 
@@ -170,8 +170,15 @@ func NewPumpsClient(etcdURLs, strategy string, timeout time.Duration, securityOp
 	newPumpsClient.Selector.SetPumps(copyPumps(newPumpsClient.Pumps.AvaliablePumps))
 
 	newPumpsClient.wg.Add(2)
-	go newPumpsClient.watchStatus(revision)
-	go newPumpsClient.detect()
+	go func() {
+		newPumpsClient.watchStatus(revision)
+		newPumpsClient.wg.Done()
+	}()
+
+	go func() {
+		newPumpsClient.detect()
+		newPumpsClient.wg.Done()
+	}()
 
 	return newPumpsClient, nil
 }
@@ -238,7 +245,7 @@ func (c *PumpsClient) getPumpStatus(pctx context.Context) (revision int64, err e
 	return revision, nil
 }
 
-// WriteBinlog writes binlog to a situable pump. Tips: will never return error for commit/rollback binlog.
+// WriteBinlog writes binlog to a suitable pump. Tips: will never return error for commit/rollback binlog.
 func (c *PumpsClient) WriteBinlog(binlog *pb.Binlog) error {
 	c.RLock()
 	pumpNum := len(c.Pumps.AvaliablePumps)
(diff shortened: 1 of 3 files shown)


tidb-binlog/pump_client/client.go
```diff
@@ -318,52 +325,59 @@ func (c *PumpsClient) WriteBinlog(binlog *pb.Binlog) error {
...
+		if !pump.ShouldBeUsable() {
 			continue
 		}
 
 		pump.ResetGrpcClient()
 
 		resp, err = pump.WriteBinlog(req, c.BinlogWriteTimeout)
-		if err == nil {
-			if resp.Errmsg != "" {
-				err = errors.New(resp.Errmsg)
```
july2993, line 356: this err ineffassign before

tidb-binlog/pump_client/client.go
```diff
@@ -318,52 +325,59 @@ func (c *PumpsClient) WriteBinlog(binlog *pb.Binlog) error {
...
 	}
 
-	unAvaliablePumps := make([]*PumpStatus, 0, 3)
+	allPumps := make([]*PumpStatus, 0, 3)
 	c.RLock()
-	for _, pump := range c.Pumps.UnAvaliablePumps {
-		unAvaliablePumps = append(unAvaliablePumps, pump)
+	for _, pump := range c.Pumps.Pumps {
+		allPumps = append(allPumps, pump)
 	}
```
suzaku: We can use `allPumps := copyPumps(c.Pumps.Pumps)` here?

tidb-binlog/pump_client/client.go
```diff
@@ -318,52 +325,59 @@ func (c *PumpsClient) WriteBinlog(binlog *pb.Binlog) error {
...
 	c.RUnlock()
 
 	var resp *pb.WriteBinlogResp
-	// send binlog to unavaliable pumps to retry again.
-	for _, pump := range unAvaliablePumps {
-		if !pump.IsUsable() {
+	// send binlog to unavailable pumps to retry again.
+	for _, pump := range allPumps {
+		if !pump.ShouldBeUsable() {
```
suzaku, line 353: `IsUsable` also considered `errNum`, is that still relevant now?

tidb-binlog/pump_client/client.go
```diff
@@ -487,13 +501,12 @@ func (c *PumpsClient) watchStatus(revision int64) {
 		case wresp := <-rch:
 			if wresp.Err() != nil {
 				// meet error, watch from the latest revision.
+				// pump will update the key periodly, it's ok for we to lost some event here
 				log.Warn("[pumps client] watch status meet error", zap.Error(wresp.Err()))
-				rch = c.EtcdRegistry.WatchNode(c.ctx, c.nodePath, revision)
+				rch = c.EtcdRegistry.WatchNode(c.ctx, c.nodePath, 0)
```
suzaku, line 503: What does it mean to set revision to 0 here?

tidb-binlog/pump_client/client.go
july2993: every online pump should be writable, we should try our best to write success in backoffcase or the service is down.
and the `errNum` is really not clear to tell the behavior here with concurrently `WriteBinlog` in client.go.
```
// IsUsable returns true if pump is usable.
func (p *PumpStatus) IsUsable() bool {
    if !p.ShouldBeUsable() {
        return false
    }

    if atomic.LoadInt64(&p.ErrNum) > defaultMaxErrNums {
        return false
    }

    return true
}
```

tidb-binlog/pump_client/client.go
july2993: well, finally it will reach here...
```
// Watch watchs the events of key with prefix.
func (e *Client) Watch(ctx context.Context, prefix string, revision int64) clientv3.WatchChan {
    if revision > 0 {
        return e.client.Watch(ctx, prefix, clientv3.WithPrefix(), clientv3.WithRev(revision))
    }
    return e.client.Watch(ctx, prefix, clientv3.WithPrefix())
}
```
it means will not specify revision to watch, then it will just return the later evens while the Watch reach etcd but not return the evens start at the specified revision.

Screen Shot 2019-08-08 at 11 29 42 AM

tidb-binlog/pump_client/client.go
suzaku: How do we decide when to call `IsUsable` and `ShouldBeUsable` in the future?

july2993: according to the code before, IsUsable() means the pump is usable really( may return false if reach some write fails although the state is online.

suzaku (commented):
LGTM

IANTHEREAL (approved):
LGTM

PR #271, opened 2019-09-24
check: update instruction for auto-increment key checking

WangXiangUSTC:
we decide to remove the column mapping, so need update the instruction

update instruction

### DIFF
--- a/pkg/check/table_structure.go
+++ b/pkg/check/table_structure.go
@@ -374,7 +374,7 @@ func (c *ShardingTablesCheck) checkAutoIncrementKey(instance, schema, table stri
 		if !hasMatchedRule {
 			r.State = StateFailure
 			r.ErrorMsg = fmt.Sprintf("instance %s table `%s`.`%s` of sharding %s have auto-increment key, would conflict with each other to cause data corruption", instance, schema, table, c.name)
-			r.Instruction = "please set column mapping rules for them, or handle it by yourself"
+			r.Instruction = "please handle it by yourself, read https://pingcap.com/docs-cn/v3.0/reference/tools/data-migration/usage-scenarios/best-practice-dm-shard/#自增主键冲突处理 for more detail"
 			r.Extra = AutoIncrementKeyChecking
 			return false
 		}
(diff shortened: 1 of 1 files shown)


pkg/check/table_structure.go
```diff
@@ -374,7 +374,7 @@ func (c *ShardingTablesCheck) checkAutoIncrementKey(instance, schema, table stri
 		if !hasMatchedRule {
 			r.State = StateFailure
 			r.ErrorMsg = fmt.Sprintf("instance %s table `%s`.`%s` of sharding %s have auto-increment key, would conflict with each other to cause data corruption", instance, schema, table, c.name)
-			r.Instruction = "please set column mapping rules for them, or handle it by yourself"
+			r.Instruction = "please handle it by yourself, read https://pingcap.com/docs-cn/dev/reference/tools/data-migration/usage-scenarios/best-practice-dm-shard/#自增主键冲突处理 for more detail"
```
IANTHEREAL: also put English document link here？

WangXiangUSTC: don't have english doc now, update it later?

IANTHEREAL: add english doc firstly, some foreign users need it

WangXiangUSTC: ok

csuzhangxc: should we link to `dev` or `v3.0`?

csuzhangxc: If the user want to _continue_ after _handled it by themself_, what should they do in _dmctl_?

WangXiangUSTC: because dm's version is independent, so I think dev is better.

WangXiangUSTC: also add a tip for how to do? but this is a pkg, infact I think add a dm's link here is not very suitable.

WangXiangUSTC:
it will take some days for i18n to translate the document to english, and I think `check` is a pkg, should not contain some information with DM. So how about just remove the tip about column mapping in `check`, and add some tip in DM? @csuzhangxc @GregoryIan

csuzhangxc:
> it will take some days for i18n to translate the document to english, and I think `check` is a pkg, should not contain some information with DM. So how about just remove the tip about column mapping in `check`, and add some tip in DM? @csuzhangxc @GregoryIan

also remove tips for other pkgs?

WangXiangUSTC:
I had checked other check item, their instruction are common, don't concern about specified project. @csuzhangxc

csuzhangxc:
> I had checked other check item, their instruction are common, don't concern about specified project. @csuzhangxc

can we add a common instruction in the pkg, but overwrite it in DM?

WangXiangUSTC:
yes, now all check items' result have common instruction. @csuzhangxc

csuzhangxc (commented):
LGTM

IANTHEREAL (approved):
LGTM

Issue #272, opened 2019-09-25
unsupported regex pattern in filter rule will cause panic


amyangfei:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
```
panic: regexp: Compile(`(?i)^t[0-9]+((?!_copy).)*$`): error parsing regexp: invalid or unsupported Perl syntax: `(?!`

goroutine 55 [running]:
regexp.MustCompile(0xc00046e2c0, 0x1a, 0x4)
        /root/.gvm/gos/go1.12/src/regexp/regexp.go:272 +0x159
github.com/pingcap/tidb-tools/pkg/filter.(*Filter).addOneRegex(0xc000472580, 0xc00046e0e0, 0x17)
        /root/.gvm/pkgsets/go1.12/global/pkg/mod/github.com/pingcap/tidb-tools@v3.0.0-beta.1.0.20190821032033-e6ccf3994944+incompatible/pkg/filter/filter.go:156 +0xba
github.com/pingcap/tidb-tools/pkg/filter.(*Filter).genRegexMap(0xc000472580)
        /root/.gvm/pkgsets/go1.12/global/pkg/mod/github.com/pingcap/tidb-tools@v3.0.0-beta.1.0.20190821032033-e6ccf3994944+incompatible/pkg/filter/filter.go:132 +0xfb
github.com/pingcap/tidb-tools/pkg/filter.New(0x1343800, 0xc000389e00, 0x155e947)
        /root/.gvm/pkgsets/go1.12/global/pkg/mod/github.com/pingcap/tidb-tools@v3.0.0-beta.1.0.20190821032033-e6ccf3994944+incompatible/pkg/filter/filter.go:117 +0xbe
github.com/pingcap/dm/checker.(*Checker).Init(0xc00043e2a0, 0x0, 0x0)
```

2. What did you expect to see?

The interface provided in `pkg/filter` returns error instead of panic.

3. What did you see instead?

panic

4. What version of TiDB are you using (`tidb-server -V` or run `select tidb_version();` on TiDB)?
github.com/pingcap/tidb-tools v3.0.0-beta.1.0.20190821032033-e6ccf3994944+incompatible

PR #273, opened 2019-09-29
filter: return error if passed invalid regex pattern

amyangfei:

Fix issue https://github.com/pingcap/tidb-tools/issues/272

Use `regexp.Compile` to replace `regexp.MustCompile`, and return the error if necessary. Note this pr will change the return list of `pkg/filter.New` 

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/filter/filter.go
+++ b/pkg/filter/filter.go
@@ -117,7 +117,10 @@ func New(caseSensitive bool, rules *Rules) (*Filter, error) {
 		items: make(map[string]ActionType),
 	}
 	err := f.genRegexMap()
-	return f, err
+	if err != nil {
+		return nil, err
+	}
+	return f, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/filter/filter.go
```diff
@@ -114,35 +116,55 @@ func New(caseSensitive bool, rules *Rules) *Filter {
 	f.c = &cache{
 		items: make(map[string]ActionType),
 	}
-	f.genRegexMap()
-	return f
+	err := f.genRegexMap()
+	return f, err
```
csuzhangxc: if `err != nil`, the behavior of `f` is unpredictable, should we need to return `nil` rather than `f`?

WangXiangUSTC (commented):
LGTM

csuzhangxc (approved):
LGTM

PR #274, opened 2019-10-09
restore-util: Implement split/scatter

5kbpers:

This PR implements the `Prepare` stage of [tikv-importer](https://github.com/tikv/importer) (called split/scatter) as a lib, which would be used by tidb-lightning and br.

Split policy:
1. split by the new prefix of rewrite rules
2. split by the end key of backup files

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects

 - Increased code complexity


kennytm:
Hold on, I don't think we're supposed to do size-based splitting inside BR, since the SSTs are already split into ideal size.

5kbpers:
> Hold on, I don't think we're supposed to do size-based splitting inside BR, since the SSTs are already split into ideal size.

'Ideal size' means the size of each SST can be considered as the size of a region?

kennytm:
@5kbpers Yes. And we're supposed to use only the range and key-rewrite rules for split and scatter, to avoid an extra O(N) reading of the SST files.

5kbpers:
> @5kbpers Yes. And we're supposed to use only the range and key-rewrite rules for split and scatter, to avoid an extra O(N) reading of the SST files.

Get it. Thank you!

5kbpers:
@kennytm I think we could do the key-rewrite outside, it only required by BR, we can pass the rewritten key to the lib.

kennytm:
@5kbpers I think this library should still be aware of the key rewrite rules, because if we have the range `[AA, AZ)`, and there is a prefix-rewrite rules `ABC ↦ XYZ, AXX ↦ XXX`, then we need to split many ranges:

1. Pre-rewrite: `[AA0, ABC)` / Post-rewrite: `[AA0, ABC)`
2. Pre-rewrite: `[ABC, ABD)` / Post-rewrite: `[XYZ, XZ0)`
3. Pre-rewrite: `[ABD, AXX)` / Post-rewrite: `[ABD, AXX)`
4. Pre-rewrite: `[AXX, AXY)` / Post-rewrite: `[XXX, XXY)`
5. Pre-rewrite: `[AXY, AZ0)` / Post-rewrite: `[AXY, AZ0)`

pkg/restore-util/client.go
```diff
@@ -0,0 +1,167 @@
...
+	GetRegionByID(ctx context.Context, regionID uint64) (*RegionInfo, error)
+	// SplitRegion splits a region from a key, if key is not included in the region, it will return nil.
+	SplitRegion(ctx context.Context, regionInfo *RegionInfo, key []byte) (*RegionInfo, error)
+	// ScatterRegion scatters a specified region.
+	ScatterRegion(ctx context.Context, regionInfo *RegionInfo) error
+	// GetOperator gets the status of operator of the specified region.
+	GetOperator(ctx context.Context, regionID uint64) (*pdpb.GetOperatorResponse, error)
+	// ScanRegion gets a list of regions, starts from the region that contains key.
+	// Limit limits the maximum number of regions returned.
+	ScanRegions(ctx context.Context, key []byte, limit int) ([]*RegionInfo, error)
```
overvenus: Looks like we do not need it since there is no referenece to the method.

go.mod1
```diff
@@ -1,25 +1,117 @@
 module github.com/pingcap/tidb-tools
 
 require (
+	cloud.google.com/go/bigquery v1.1.0 // indirect
+	cloud.google.com/go/storage v1.1.0 // indirect
```
kennytm: Run `go mod tidy` to reduce the number of unnecessary entries.

pkg/restore-util/range.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// InsertRanges inserts ranges into the range tree.
+// it returns true if all ranges inserted successfully.
+// it returns false if there are some overlapped ranges.
+func (rt *RangeTree) InsertRange(rg Range) bool {
+	if rt.tree.ReplaceOrInsert(&rg) != nil {
+		return false
+	}
+	return true
```
kennytm:
```go
	return rt.tree.ReplaceOrInsert(&rg) == nil
```

pkg/restore-util/range.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// RangeIterator allows callers of Ascend to iterate in-order over portions of
+// the tree. When this function returns false, iteration will stop and the
+// associated Ascend function will immediately return.
+type RangeIterator func(rg *Range) bool
+
+// Ascend calls the iterator for every value in the tree within [first, last],
+// util the iterator returns false.
```
kennytm:
```suggestion
// until the iterator returns false.
```

pkg/restore-util/split.go
```diff
@@ -0,0 +1,220 @@
...
+	"time"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/kvproto/pkg/import_sstpb"
+	"github.com/pingcap/kvproto/pkg/pdpb"
+)
+
+var SplitWaitMaxRetryTimes = 64
+var SplitWaitIntervalMillis = 8
+var SplitMaxWaitIntervalMillis = 1000
```
kennytm: I suggest you make these of type `time.Time` e.g.

```go
var SplitWaitInterval = 8 * time.Millisecond
var SplitMaxWaitInterval = time.Second
```

BTW can these be `const` instead of `var`?

pkg/restore-util/split.go
```diff
@@ -0,0 +1,220 @@
...
+
+var ScatterWaitMaxRetryTimes = 128
+var ScatterWaitIntervalMillis = 50
+var ScatterMaxWaitIntervalMillis = 5000
```
kennytm: (ditto)

pkg/restore-util/split.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+}
+
+// Split executes a region split. It will split regions by the rewrite rules,
+// then it will split regions by the end key of each range.
+func (rs *RegionSplitter) Split(ctx context.Context, ranges []Range, rules []*import_sstpb.RewriteRule) error {
+	var wg sync.WaitGroup
+	rangeTree, ok := newRangeTreeWithRewrite(ranges, rules)
+	if !ok {
+		return errors.Errorf("ranges overlapped")
```
```suggestion
		return errors.New("ranges overlapped")
```

pkg/restore-util/split.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	rangeTree.Ascend(func(rg *Range) bool {
+		if rg == nil {
+			return false
+		}
+		err = rs.maybeSplitRegion(ctx, rg, &wg)
+		if err != nil {
+			return false
+		}
+		return true
```
```go
return err == nil
```

pkg/restore-util/split.go
```diff
@@ -0,0 +1,220 @@
...
+
+func (rs *RegionSplitter) hasRegion(ctx context.Context, regionID uint64) (bool, error) {
+	regionInfo, err := rs.client.GetRegionByID(ctx, regionID)
+	if err != nil {
+		return false, err
+	}
+	if regionInfo == nil {
+		return false, nil
+	}
+	return true, nil
```
```go
return regionInfo != nil, nil
```

pkg/restore-util/split.go
```diff
@@ -0,0 +1,220 @@
...
+
+func (rs *RegionSplitter) waitForScatter(ctx context.Context, wg *sync.WaitGroup, regionID uint64) {
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		interval := ScatterWaitIntervalMillis
+		for i := 0; i < ScatterWaitMaxRetryTimes; i++ {
+			ok, err := rs.hasRegion(ctx, regionID)
+			if err != nil {
+				return
```
kennytm, line 156: log the error

pkg/restore-util/split.go
```diff
@@ -0,0 +1,220 @@
...
+	return beforeEnd(splitKey, regionInfo.Region.GetEndKey())
+}
+
+func beforeEnd(key []byte, end []byte) bool {
+	return bytes.Compare(key, end) < 0 || len(end) == 0
+}
+
+func replacePrefix(s []byte, oldPrefix []byte, newPrefix []byte) []byte {
+	if bytes.HasPrefix(s, oldPrefix) {
+		return append(newPrefix, bytes.TrimPrefix(s, oldPrefix)...)
```
```suggestion
		return append(newPrefix, s[len(oldPrefix):]...)
```

pkg/restore-util/client.go
```diff
@@ -0,0 +1,167 @@
...
+	storeID := peer.GetStoreId()
+	store, err := c.GetStore(ctx, storeID)
+	if err != nil {
+		return nil, err
+	}
+	req := &kvrpcpb.SplitRegionRequest{
+		Context:   reqCtx,
+		SplitKeys: [][]byte{key},
+	}
+	conn, err := grpc.Dial(store.GetAddress(), grpc.WithInsecure())
```
kennytm, line 111: Handle this error (or make it `_` if this can be ignored).

pkg/restore-util/client.go
```diff
@@ -0,0 +1,167 @@
...
+	if resp.RegionError != nil {
+		return nil, errors.Errorf("split region %d failed, got region error: %v", regionInfo.Region.GetId(), resp.RegionError)
+	}
+
+	regions := resp.GetRegions()
+	var newRegion *metapb.Region
+	for _, r := range regions {
+		// Assume the new region is the left one.
+		if bytes.Compare(r.GetStartKey(), regionInfo.Region.GetStartKey()) == 0 {
+			newRegion = r
```
kennytm, line 136: Should we `break` here?

pkg/restore-util/client.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	if resp.RegionError != nil {
+		return nil, errors.Errorf("split region %d failed, got region error: %v", regionInfo.Region.GetId(), resp.RegionError)
+	}
+
+	regions := resp.GetRegions()
+	var newRegion *metapb.Region
+	for _, r := range regions {
+		// Assume the new region is the left one.
+		if bytes.Compare(r.GetStartKey(), regionInfo.Region.GetStartKey()) == 0 {
```
kennytm:
```suggestion
		if bytes.Equal(r.GetStartKey(), regionInfo.Region.GetStartKey()) {
```

pkg/restore-util/client.go
```diff
@@ -0,0 +1,167 @@
...
+			newRegion = r
+		}
+	}
+	if newRegion == nil {
+		return nil, errors.Errorf("split region failed")
```
kennytm:
```suggestion
		return nil, errors.New("split region failed")
```

pkg/restore-util/client.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	var leader *metapb.Peer
+	// Assume the leaders will be at the same store.
+	if regionInfo.Leader != nil {
+		for _, p := range newRegion.GetPeers() {
+			if p.GetStoreId() == regionInfo.Leader.GetStoreId() {
+				leader = p
```
kennytm, line 148: Should we `break` here?

pkg/restore-util/client.go
```diff
@@ -0,0 +1,167 @@
...
+		RegionEpoch: regionInfo.Region.RegionEpoch,
+		Peer:        peer,
+	}
+	storeID := peer.GetStoreId()
+	store, err := c.GetStore(ctx, storeID)
+	if err != nil {
+		return nil, err
+	}
+	req := &kvrpcpb.SplitRegionRequest{
+		Context:   reqCtx,
```
kennytm: Since `reqCtx` is only used once I prefer inlining that struct literal here 🙂 (or maybe even inline `req` into line 112).

```suggestion
		Context:   &kvrpcpb.Context{
			RegionId:    regionInfo.Region.Id,
			RegionEpoch: regionInfo.Region.RegionEpoch,
			Peer:        peer,
		},
```

pkg/restore-util/range_test.go
```diff
@@ -0,0 +1,54 @@
...
+			EndKey:   rg.EndKey,
+		})
+		return true
+	})
+	if len(ranges) != len(resRanges) {
+		t.Fatalf("some inserted ranges missed: rg=%v res=%v", ranges, resRanges)
+	}
+	for i, rg := range ranges {
+		res := resRanges[i]
+		if bytes.Compare(rg.StartKey, res.StartKey) != 0 || bytes.Compare(rg.EndKey, res.EndKey) != 0 {
```
kennytm:
```suggestion
		if !bytes.Equal(rg.StartKey, res.StartKey) || !bytes.Equal(rg.EndKey, res.EndKey) {
```

pkg/restore-util/split_test.go
```diff
@@ -0,0 +1,213 @@
...
+// 		[, aay), [aay, bb), [bb, bba), [bba, bbf), [bbf, bbh), [bbh, cca), [cca, xx), [xx, xxe), [xxe, xxz), [xxz, )
+func validateRegions(regions map[uint64]*RegionInfo) bool {
+	keys := [11]string{"", "aay", "bb", "bba", "bbf", "bbh", "cca", "xx", "xxe", "xxz", ""}
+	if len(regions) != 10 {
+		return false
+	}
+	for i := 1; i < 11; i++ {
+		var found bool
+		for _, region := range regions {
+			if bytes.Compare(region.Region.GetStartKey(), []byte(keys[i-1])) == 0 &&
```
kennytm:
```suggestion
			if bytes.Equal(region.Region.GetStartKey(), []byte(keys[i-1])) &&
```

pkg/restore-util/split_test.go
```diff
@@ -0,0 +1,213 @@
...
+				bytes.Compare(region.Region.GetEndKey(), []byte(keys[i])) == 0 {
```
kennytm:
```suggestion
				bytes.Equal(region.Region.GetEndKey(), []byte(keys[i])) {
```

pkg/restore-util/split_test.go
```diff
@@ -0,0 +1,213 @@
...
+				found = true
```
kennytm: Could use a `continue outer` here

kennytm (commented):
Rest LGTM.

We need to determine whether the rewrite rules act on *origin* key (no `z`) or *data* key (with `z`). If latter we need to strip the leading `z`.

pkg/restore-util/client.go
5kbpers: we will use it in BR

5kbpers:
PTAL @kennytm @overvenus

5kbpers:
PTAL

kennytm (commented):
LGTM

overvenus (approved):
:+1:

PR #277, opened 2019-10-30
restore-util: add split retry interval

5kbpers:

This PR adds split retry interval, to reduce the failure probability of splitting a region.
And it resets some dependencies to release-3.1 branch.

# What is changed and how it works?

### Check List 

Tests 

 - Integration test


overvenus (approved):
LGTM

PR #279, opened 2019-11-01
restore-util: wait for scatter region sequentially 

5kbpers:

Waiting for scatter region sequentially to avoid to create goroutine by RegionSplitter

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/pkg/restore-util/client.go
+++ b/pkg/restore-util/client.go
@@ -109,6 +109,7 @@ func (c *pdClient) SplitRegion(ctx context.Context, regionInfo *RegionInfo, key 
 		return nil, err
 	}
 	conn, err := grpc.Dial(store.GetAddress(), grpc.WithInsecure())
+	defer conn.Close()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


pkg/restore-util/split.go
```diff
@@ -46,43 +45,54 @@ func NewRegionSplitter(client Client) *RegionSplitter {
...
 		}
-		err = rs.maybeSplitRegion(ctx, rg, &wg)
+		var newRegion *RegionInfo
+		newRegion, err = rs.maybeSplitRegion(ctx, rg)
+		if err == nil {
+			if newRegion != nil {
+				scatterRegions = append(scatterRegions, newRegion)
+			}
+		}
 		return err == nil
```
```go
if err != nil {
    return false
}
if newRegion != nil {
    scatterRegions = append(scatterRegions, newRegion)
}
return true
```

pkg/restore-util/client.go
```diff
@@ -109,6 +109,7 @@ func (c *pdClient) SplitRegion(ctx context.Context, regionInfo *RegionInfo, key
 		return nil, err
 	}
 	conn, err := grpc.Dial(store.GetAddress(), grpc.WithInsecure())
+	defer conn.Close()
```
overvenus, line 112: Nice catch.

Also, as a future improvement, we can cache the gRPC connection in the pdClient.

pkg/restore-util/split.go
5kbpers: oh, I forget to modify it after removing debug log.😂️

kennytm (commented):
LGTM

overvenus (approved):
LGTM

PR #281, opened 2019-11-05
restore-util: add on split hook

overvenus:

Add a hook that will be called before splitting a range.

### Check List 

Tests 

 - Unit test

Code changes

### DIFF
--- a/pkg/restore-util/split.go
+++ b/pkg/restore-util/split.go
@@ -66,7 +66,9 @@ func (rs *RegionSplitter) Split(
 		if rg == nil {
 			return false
 		}
-		onSplit(rg)
+		if onSplit != nil {
+			onSplit(rg)
+		}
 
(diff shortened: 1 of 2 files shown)


pkg/restore-util/split.go
```diff
@@ -57,6 +66,8 @@ func (rs *RegionSplitter) Split(ctx context.Context, ranges []Range, rewriteRule
 		if rg == nil {
 			return false
 		}
+		onSplit(rg)
```
```suggestion
		if onSplit != nil {
			onSplit(rg)
		}
```

kennytm (commented):
Rest LGTM

pkg/restore-util/split.go
overvenus: Thanks!

5kbpers (approved):
LGTM

PR #283, opened 2019-11-06
restore-util: fix returned new region is nil

5kbpers:

In TiKV 3.1, the `Regions` field of `SplitRegion` response is usually empty, then the returned new region is always nil in region-util, so that waiting for scattering regions does not work

Check the `Left` field of `SplitRegion` response.

### Check List 

Tests 

 - Integration test


overvenus (dismissed):
:shipit:

pkg/restore-util/split.go
```diff
@@ -83,10 +84,14 @@ func (rs *RegionSplitter) Split(
 	if err != nil {
 		return errors.Trace(err)
 	}
-
+	log.Info("splitting regions done, wait for scattering regions",
+		zap.Int("regions", len(scatterRegions)), zap.Int64("cost_seconds", int64(time.Since(startTime)/time.Second)))
```
```suggestion
		zap.Int("regions", len(scatterRegions)), zap.Duration("cost", time.Since(startTime)))
```

pkg/restore-util/split.go
```diff
@@ -83,10 +84,14 @@ func (rs *RegionSplitter) Split(
...
+	startTime = time.Now()
 	for _, region := range scatterRegions {
 		rs.waitForScatterRegion(ctx, region)
 	}
+	log.Info("waiting for scattering regions done",
+		zap.Int("regions", len(scatterRegions)), zap.Int64("cost_seconds", int64(time.Since(startTime)/time.Second)))
```
```suggestion
		zap.Int("regions", len(scatterRegions)), zap.Duration("cost", time.Since(startTime)))
```

kennytm (commented):
Rest LGTM (after fixing unit test)

5kbpers:
Seems requires to update the tidb dependency, wait for https://github.com/pingcap/tidb-tools/pull/284

overvenus (approved):

PR #286, opened 2019-11-18
pkg/etcd: return revision when create key

WangXiangUSTC:
return revision when creating a key. used for dm

### Check List 

Tests 

 - Unit test

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -90,7 +90,7 @@ func (e *Client) Create(ctx context.Context, key string, val string, opts []clie
 	}
 
 	// impossible to happen
-	return -1, nil
+	return 0, errors.New("revision is unknown")
 }
 
 // Get returns a key/value matchs the given key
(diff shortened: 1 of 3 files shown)


pkg/etcd/etcd.go
```diff
@@ -70,22 +70,27 @@ func (e *Client) Close() error {
...
+		return 0, errors.AlreadyExistsf("key %s in etcd", key)
 	}
 
-	return nil
+	if txnResp.Header != nil {
+		return txnResp.Header.Revision, nil
+	}
+
+	// impossible to happen
+	return -1, nil
```
csuzhangxc: How about returning a new error?

csuzhangxc (commented):
rest LGTM

pkg/etcd/etcd.go
WangXiangUSTC: updated https://github.com/pingcap/docs-cn/pull/2037

csuzhangxc (commented):
LGTM

lichunzhu (approved):
LGTM

PR #293, opened 2019-11-26
restore-util: fix overlapped error message

5kbpers:

The error message is always too long.

Make the error message more friendly.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

Side effects

Related changes

N/A

### DIFF
--- a/pkg/restore-util/client.go
+++ b/pkg/restore-util/client.go
@@ -3,7 +3,6 @@ package restore_util
 import (
 	"bytes"
 	"context"
-	"fmt"
 	"sync"
 
 	"github.com/pingcap/errors"
@@ -100,7 +99,7 @@ func (c *pdClient) SplitRegion(ctx context.Context, regionInfo *RegionInfo, key 
 		peer = regionInfo.Leader
 	} else {
 		if len(regionInfo.Region.Peers) == 0 {
-			return nil, errors.NewNoStackError("region does not have peer")
+			return nil, errors.New("region does not have peer")
 		}
 		peer = regionInfo.Region.Peers[0]
 	}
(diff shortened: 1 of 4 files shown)


pkg/restore-util/client.go
```diff
@@ -99,7 +100,7 @@ func (c *pdClient) SplitRegion(ctx context.Context, regionInfo *RegionInfo, key
 		peer = regionInfo.Leader
 	} else {
 		if len(regionInfo.Region.Peers) == 0 {
-			return nil, errors.New("region does not have peer")
+			return nil, fmt.Errorf("region does not have peer")
```
kennytm: why changing these?

overvenus: What about `NewNoStackError`?

https://godoc.org/github.com/pingcap/errors#NewNoStackError

5kbpers: I think this error trace is not necessary.

5kbpers: > What about `NewNoStackError`?
> 
> https://godoc.org/github.com/pingcap/errors#NewNoStackError

Seems it does not support error format.

5kbpers:
Comments addressed, PTAL @kennytm @overvenus

overvenus (approved):
LGTM :shipit:

pkg/restore-util/client.go
kennytm: Well I find it strange why we need to specifically use the no-stack version, since the `String()` of the error won't include the stack.

5kbpers: > Well I find it strange why we need to specifically use the no-stack version, since the `String()` of the error won't include the stack.

yes, I have changed back it to the with-stack version.

5kbpers:
PTAL @kennytm

kennytm (approved):

PR #296, opened 2019-11-29
pkg/etcd: support do some operations in one transaction

WangXiangUSTC:

dm-master embed etcd, and need to do some operations in one transaction.

support do some etcd operations in one transaction

### Check List 

Tests 

 - Unit test

Code changes

### DIFF
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -258,6 +258,10 @@ func (e *Client) DoTxn(ctx context.Context, operations []*Operation) (int64, err
 		operation.Key = keyWithPrefix(e.rootPath, operation.Key)
 
 		if operation.TTL > 0 {
+			if operation.Tp == DeleteOp {
+				return 0, errors.Errorf("unexpected TTL in delete operation")
+			}
+
 			lcr, err := e.client.Lease.Grant(ctx, operation.TTL)
 			if err != nil {
 				return 0, errors.Trace(err)
--- a/pkg/etcd/etcd.go
+++ b/pkg/etcd/etcd.go
@@ -30,6 +30,31 @@ type Node struct {
 	Childs map[string]*Node
 }
 
+// OpType is operation's type in etcd
+type OpType string
+
+var (
+	// CreateOp is create operation type
+	CreateOp OpType = "create"
+
+	// UpdateOp is update operation type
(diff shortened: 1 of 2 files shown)


pkg/etcd/etcd.go
```diff
@@ -217,6 +248,57 @@ func (e *Client) Watch(ctx context.Context, prefix string, revision int64) clien
...
+// DoTxn does some operation in one transaction.
+// Note: should only have one opereration for one key, otherwise will get duplicate key error.
+func (e *Client) DoTxn(ctx context.Context, operations []*Operation) (int64, error) {
+	cmps := make([]clientv3.Cmp, 0, len(operations))
+	ops := make([]clientv3.Op, 0, len(operations))
+
+	for _, operation := range operations {
+		operation.Key = keyWithPrefix(e.rootPath, operation.Key)
+
+		if operation.TTL > 0 {
```
csuzhangxc, line 260: should we need to check `operation.Tp != DeleteOp` here because `WithLease` only takes effect for `PUT` operation?

pkg/etcd/etcd_test.go
```diff
@@ -193,6 +193,134 @@ func (t *testEtcdSuite) TestDelete(c *C) {
...
+	c.Assert(err, ErrorMatches, ".* not found")
+
+	// case5: do transaction failed because can't operate one key in one transaction
+	ops = []*Operation{
+		{
+			Tp:    CreateOp,
+			Key:   "test6",
+			Value: "6",
+		}, {
+			Tp:    UpdateOp,
```
csuzhangxc, line 311: How about adding a case for `create` on an existing key and a case for `delete` on a non-existing key?

csuzhangxc (commented):
rest LGTM

pkg/etcd/etcd.go
WangXiangUSTC: yes, if set TTL for delete,  etcd client will panic `Panic: unexpected lease in delete`, I add check in af2a844

pkg/etcd/etcd_test.go
WangXiangUSTC: add case in af2a844

csuzhangxc (commented):
LGTM

amyangfei (approved):
LGTM

PR #302, opened 2019-12-11
table-rule-selector: add [...] and [!...] wildcard matching

lzmhhh123:
Resolve the first two tasks for https://github.com/pingcap/dm/issues/404

add [...] and [!...] wildcard matching

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity


pkg/table-rule-selector/selector_test.go
```diff
@@ -52,6 +55,9 @@ var _ = Suite(&testSelectorSuite{
 		{"schema1", "test1", 2, []string{"schema*", "", "schema*", "test*"}},
 		{"t1", "test1", 1, []string{"t*", "test*"}},
 		{"schema1", "abc1", 2, []string{"schema*", "", "schema*", "abc*"}},
+		{"ikj", "ikb", 1, []string{"ik[hjkl]", "ik[!zxc]"}},
+		{"ikh", "iky", 2, []string{"ik[hjkl]", "ik[!zxc]", "ik[f-h]", "ik[!a-ce-g]"}},
```
amyangfei, line 66: It seems this case still has random failure, maybe caused by the unordered map iteration

lzmhhh123: Yes, I'll change a better way to compare for results in the test.

lzmhhh123: Done.

amyangfei:
PTAL @csuzhangxc

pkg/table-rule-selector/trie_selector.go
```diff
@@ -130,62 +229,116 @@ func (t *trieSelector) insertTable(schema, table string, rule interface{}, repla
...
+	}
+	item := &rangeItem{}
+	startI := 1
+	if pattern[startI] == rangeNot {
+		startI++
+		item.hasNot = true
+	}
+	for i := startI; i < nextI; i++ {
+		if pattern[i+1] == rangeBetween {
+			item.ranges = append(item.ranges, ran{pattern[i], pattern[i+2], true})
```
amyangfei, line 267: `ran{pattern[i], pattern[i=2]` does not handle un-paired between such as `"ik[A-]"` very well, which will be recognized as `ik[A-]]`
Maybe we can make regulation about how the range selector works in some corner case, such as `[]` matches nothing, `[` matches `[`, `[!]` matches everything, `[a-]` is not determined

pkg/table-rule-selector/trie_selector.go
```diff
@@ -76,20 +84,111 @@ type trieSelector struct {
...
+}
+
+type rangeItem struct {
+	*baseItem
+
+	hasNot bool
+	ranges []ran
+}
+
+func (i *rangeItem) match(i2 *rangeItem) bool {
```
amyangfei, line 154: This function is only used to check whether `rangeItem`s are equivalent, by checking a covers b and b covers a, is it better to extract to a function like `equals`. Besides can we sort the `ran` in `rangeItem` and compare the `start-end` directly

pkg/table-rule-selector/trie_selector.go
lzmhhh123: Is `ik[A-]` a correct range? I just try it in "vim". It was recognized as `ik[A-A\--\-]`.

pkg/table-rule-selector/trie_selector.go
lzmhhh123: OK~

pkg/table-rule-selector/trie_selector.go
lzmhhh123: And `[!]` is recognized as `[\!-\!]`.

lzmhhh123: Oh, sorry, I misunderstood your mean. I'll change by your advice.

pkg/table-rule-selector/trie_selector.go
lzmhhh123: I prefer the original `match` function and add an `equal` function to call the`match` function twice. Because here are some situations like `[a-c] == [a-ca-c]`.

amyangfei (commented):
We can add more corner cases in unit test, and we can add document about the usage of wildcard in another PR (as there are various wildcard regulations, we should tell users what is support in this lib).

Rest LGTM

amyangfei (commented):
LGTM

pkg/table-rule-selector/trie_selector.go
```diff
@@ -301,14 +463,37 @@ func (t *trieSelector) track(n *node, pattern string) ([]*item, error) {
...
+				for i := range n.rItems {
+					if n.rItems[i].(*rangeItem).equal(rItem) {
+						matchIdx = i
+						break
+					}
+				}
+				if matchIdx == -1 {
+					return nil, errors.NotFoundf("pattern %v", pattern)
+				}
+				items = append(items, n.rItems[i])
```
IANTHEREAL: using `i` as index is wrong

pkg/table-rule-selector/trie_selector.go
```diff
@@ -301,14 +463,37 @@ func (t *trieSelector) track(n *node, pattern string) ([]*item, error) {
...
+				n = n.rItems[i].child()
+			}
```
IANTHEREAL, line 490: forget to add `nextI ` into `i`

pkg/table-rule-selector/selector_test.go
```diff
@@ -52,6 +62,12 @@ var _ = Suite(&testSelectorSuite{
...
+		{"iz3", "ixz", 2, []string{"i[x-z][1-3]", "i?[x-z]", "i[x-z][1-3]", "ix*"}},
+		{"!", "-", 2, []string{"[!]", "[a-]", "[!]", "[a-c-f]"}},
+		{"!", "c", 1, []string{"[!]", "[a-c-f]"}},
+		{"d", "zxcv", 1, []string{"[!a-c!f-g]", "*"}},
 	},
 	removeCases: []string{"schema*", "", "a?c", "t2_ab*"},
```
IANTHEREAL: please add case to test `Remove` function

IANTHEREAL (approved):
LGTM

PR #303, opened 2019-12-11
*: update tidb

disksing:
tidb dependency is too old. Need to update.

- update modules.
- migrate ddl updates

### Check List 

Tests 
 - Unit test


WangXiangUSTC:
LGTM

pkg/dbutil/table.go
```diff
@@ -52,202 +50,30 @@ func GetTableInfoBySQL(createTableSQL string, sqlMode string) (table *model.Tabl
...
-				// FIXME: use a default string now to make col.IsGenerated() return true, will use real generated expr string later
-				col.GeneratedExprString = "Generated"
-				// do nothing
-			case ast.ColumnOptionFulltext:
-				// do nothing
+		// put primary key in indices
+		if table.PKIsHandle {
+			pkIndex := &model.IndexInfo{
+				Name:    model.NewCIStr("PRIMARY"),
+				Primary: true,
```
zimulala, line 62: Do we need to set the values of `State` and `Unique`?

WangXiangUSTC: set in 17be9e7

zimulala (approved):
LGTM

PR #305, opened 2019-12-20
restore-util: add upper bound time for waiting for scatter

5kbpers:
Some scatter operators always run to timeout in PD. 

Add upper bound time for waiting for scatter

### Check List 

Tests 

Side effects

### DIFF
--- a/pkg/restore-util/client.go
+++ b/pkg/restore-util/client.go
@@ -91,7 +91,7 @@ func (c *pdClient) GetRegion(ctx context.Context, key []byte) (*RegionInfo, erro
 		return nil, err
 	}
 	if region == nil {
-		return nil, errors.Errorf("region not found: key=%x", key)
+		return nil, nil
 	}
 	return &RegionInfo{
 		Region: region,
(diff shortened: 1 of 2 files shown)


pkg/restore-util/client.go
```diff
@@ -217,24 +217,24 @@ func (c *pdClient) BatchSplitRegions(ctx context.Context, regionInfo *RegionInfo
...
+			for _, p := range regions[i].GetPeers() {
 				if p.GetStoreId() == regionInfo.Leader.GetStoreId() {
 					leader = p
 					break
 				}
 			}
 		}
 		newRegionInfos = append(newRegionInfos, &RegionInfo{
-			Region: region,
+			Region: regions[i],
```
kennytm: I see no reason why these (lines 220-237) are changed (`i` is not used).

5kbpers: For test, I will change it back.

5kbpers:
PTAL, thx! @kennytm @overvenus @3pointer

overvenus (approved):
:shipit:

PR #306, opened 2019-12-23
restore-util: fix wrong url

disksing:
Signed-off-by: disksing <i@disksing.com>

Generated URL `http:/127.0.0.1...` is wrong. The reason is `path.Join` replaces `//` to `/`.

Not join prefix.

### Check List 

Tests 
manually tested with br.


lhy1024 (approved):

overvenus (approved):
LGTM

PR #307, opened 2019-12-23
binlog: add utc_time_zone field in binlog pb

zier-one:
add a field to mark if the timestamp value is in UTC
 See: pingcap/tidb-binlog#841


zier-one:
PTAL @july2993 @suzaku

tidb-binlog/slave_binlog_proto/proto/binlog.proto
```diff
@@ -88,4 +88,6 @@ message Binlog {
   // dml_data is marshalled from DML type
   optional DMLData dml_data = 3;
   optional DDLData ddl_data = 4;
+  // utc_time_zone marks if the time zone is UTC
+  optional bool          utc_time_zone  = 5;
```
kennytm: Looks strange for this to be a `bool`. There are thousands of timezones that are not UTC.

zier-one: But the only information we have is whether the timestamp is in UTC. This field is used to maintain backward compatibility.
For the old version drainer, the drainer parse time in the local time zone, and send binlog to kafka (or write file) without time zone info. Unfortunately, we can't change the behavior of the old drainer, so we lost the time zone info. 
For the new version drainer, the drainer parse time in UTC, and marks the binlog(`utc_time_zone=true`) to tell arbiter(or reparo) whether the timestamp is in UTC. So arbiter(or reparo) can handle two versions of binlog.

july2993: If the new version sets the value as UTC +0 as you said and set utc_time_zone as true, this **IS NOT** backward compatibility.because the user must learn this and upgrade their application first(like arbiter).

you and a field to specify the time zone deterministic, like

```
optional string time_zone = 5;  // set it the local time zone of drainer
```

kennytm (commented):
LGTM

july2993 (dismissed):

july2993 (approved):
better add an example, and what version will support this

PR #308, opened 2019-12-26
pkg/filter: Add a Match method for checking if a table should be kept

suzaku:

In many use cases of `pkg/filter`, we don't really want the `ApplyOn` method.
Instead, we often need to check tables or schemas one by one and see whether they should be kept or removed.

Add a `Match` method.

### Check List 

Tests 

 - Unit test


kennytm (commented):
LGTM

amyangfei (approved):
LGTM

PR #309, opened 2019-12-26
.*: update tidb version

WangXiangUSTC:

https://github.com/pingcap/tidb-tools/blob/6ca94f01c59b038d6f88bd670a8d2bca4625f8fb/pkg/dbutil/table.go#L53
this line may cause panic, and is fixed in  https://github.com/pingcap/tidb/pull/14184

update tidb's version


kennytm:
I think it's time we remove that `replace github.com/google/pprof => github.com/lonng/pprof` line from go.mod.

WangXiangUSTC:
@kennytm removed, PTAL again

kennytm (commented):
Lgtm

csuzhangxc (approved):
LGTM

PR #311, opened 2019-12-27
pkg/schemacmp: added schema compare and join routines

kennytm:

Added the `schemacmp` package for comparing and merging schema for use in DM for optimistic sharded DDL merging.

The package does not support "diffing" yet, so no way to regenerate DDL from two schemas yet.

Added the `schemacmp` package 

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes


kennytm:
(Blocked by #309.)

pkg/schemacmp/lattice.go
```diff
@@ -0,0 +1,704 @@
...
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return latticeMap{LatticeMap: result}, nil
+}
+
+// Map wraps a LatticeMap instance into a Lattice.
+func Map(lm LatticeMap) latticeMap {
```
csuzhangxc: Exported function (`Map`) with unexported return type (`latticeMap`).

pkg/schemacmp/lattice.go
```diff
@@ -0,0 +1,704 @@
...
+		}
+	}
+	return result, nil
+}
+
+// CombineCompareResult combines two comparison results.
+func CombineCompareResult(x int, y int) (int, error) {
+	switch {
+	case x == y || y == 0:
+		return x, nil
```
csuzhangxc, line 430: If `x == -1` and `y == 0` then return `-1`, but ``` we define `a < b` iff `a[i] < b[i]` for all `i` ```, is this accurate?

pkg/schemacmp/lattice_test.go
```diff
@@ -0,0 +1,329 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schemacmp_test
+
+import (
+	"bytes"
+	. "github.com/pingcap/tidb-tools/pkg/schemacmp"
+
+	. "github.com/pingcap/check"
```
csuzhangxc: Do we need to refine the order?

pkg/schemacmp/lib.go
```diff
@@ -0,0 +1,14 @@
+// Copyright 2019 PingCAP, Inc.
```
csuzhangxc: Can we remove _lib.go_ and _lib_test.go_ (and move `check.TestingT(t)` to another file)?

pkg/schemacmp/table.go
```diff
@@ -0,0 +1,417 @@
...
+)
+
+// encodeColumnInfoToLattice collects the necessary information for comparing a column.
+func encodeColumnInfoToLattice(ci *model.ColumnInfo) Tuple {
+	return Tuple{
+		MaybeSingletonInterface(ci.DefaultValue),
+		Singleton(ci.GeneratedExprString),
+		Singleton(ci.GeneratedStored),
+		Type(&ci.FieldType),
+		MaybeSingletonString(ci.Comment),
```
csuzhangxc: Is it necessary to compare the _comment_ of the column?

pkg/schemacmp/table.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+}
+
+// restoreColumnInfoFromLattice restores the text representation of a column.
+func restoreColumnInfoFromUnwrapped(col []interface{}, colName string, ctx *format.RestoreCtx) {
```
csuzhangxc: `restoreColumnInfoFromLattice` in the comment mismatch with the function name (`restoreColumnInfoFromUnwrapped`).

pkg/schemacmp/table.go
```diff
@@ -0,0 +1,417 @@
...
+	return Tuple{
+		EqualitySingleton(indexColumns),
+		Bool(!ii.Unique),
+		Bool(!ii.Primary),
+		MaybeSingletonString(ii.Comment),
+		Singleton(ii.Tp),
+	}
+}
+
+func restoreIndexIntoFromUnwrapped(index []interface{}, keyName string, ctx *format.RestoreCtx) {
```
csuzhangxc:
```suggestion
func restoreIndexInfoFromUnwrapped(index []interface{}, keyName string, ctx *format.RestoreCtx) {
```

pkg/schemacmp/table.go
```diff
@@ -0,0 +1,417 @@
...
+	if comment, ok := col[columnInfoTupleIndexComment].(string); ok && len(comment) != 0 {
+		ctx.WriteKeyWord(" COMMENT ")
+		ctx.WriteString(comment)
+	}
+}
+
+const (
+	indexInfoTupleIndexColumns = iota
+	indexInfoTupleIndexNotUnique
+	indexInfoTupleIndexNotPrimary
```
csuzhangxc, line 72: Why do we need `Not` for these two items?

pkg/schemacmp/table.go
```diff
@@ -0,0 +1,417 @@
...
+	for key, value := range a {
+		if err := f(key, value); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (columnMap) CompareWithNil(value Lattice) (int, error) {
+	if value.(Tuple)[columnInfoTupleIndexFieldTypes].(typ).hasDefault() {
```
csuzhangxc, line 171: Do we need to check `value.(Tuple)` is `ok` (the same with `columnMap.JoinWithNil`)?

pkg/schemacmp/table.go
```diff
@@ -0,0 +1,417 @@
...
+
+func Encode(ti *model.TableInfo) Table {
+	return Table{value: encodeTableInfoToLattice(ti)}
+}
+
+func (t Table) Restore(tableName string, ctx *format.RestoreCtx) {
+	restoreTableInfoFromUnwrapped(t.value.Unwrap().([]interface{}), tableName, ctx)
+}
+
+func (t Table) Compare(other Table) (int, error) {
```
csuzhangxc, line 348: How about adding some test cases for it?

pkg/schemacmp/lattice_test.go
```diff
@@ -0,0 +1,329 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schemacmp_test
```
csuzhangxc, line 14: Why do we give it another name rather than `schemacmp`?

pkg/schemacmp/table_test.go
```diff
@@ -0,0 +1,345 @@
...
+	createStmtNode, ok := node.(*ast.CreateTableStmt)
+	if !ok {
+		return nil, errors.New("not a create table statement")
+	}
+	return ddl.MockTableInfo(s.sctx, createStmtNode, 1)
+}
+
+func (s *tableSchema) TestJoinSchemas(c *C) {
+	testCases := []struct {
+		name    string
```
csuzhangxc, line 63: Is this field only used for debugging?

pkg/schemacmp/table_test.go
```diff
@@ -0,0 +1,345 @@
...
+			name:    "DM_053",
+			a:       "CREATE TABLE tb1 (c BIGINT, b VARCHAR(10))",
+			b:       "CREATE TABLE tb2 (c DOUBLE, b VARCHAR(10))",
+			joinErr: `.*"c".*distinct singletons.*`,
+		},
+		// {
+		// 	name: "DM_055",
+		// 	a:    "CREATE TABLE tb1 (a INT, b VARCHAR(10))",
+		// 	b:    "CREATE TABLE tb2 (a BIGINT, b VARCHAR(10))",
+		// 	join: "CREATE TABLE tb2 (a BIGINT, b VARCHAR(10))",
```
csuzhangxc, line 243: Do you want to support this in the future?

pkg/schemacmp/lattice.go
kennytm: Changed the comment to say `a <= b` iff `a[i] <= b[i]`.

pkg/schemacmp/lattice_test.go
kennytm: don't understand what do you mean 😅

pkg/schemacmp/table.go
kennytm: Because the fields stored `!index.Unique` instead of `index.Unique`.

pkg/schemacmp/lattice_test.go
kennytm: To ensure the test package cannot access the private members of `schemacmp`.

pkg/schemacmp/table_test.go
kennytm: Yes

pkg/schemacmp/table_test.go
kennytm: Yes. Commented.

pkg/schemacmp/lattice_test.go
csuzhangxc: 😅put the `imports` into two (or three) sections, builtin first, then others, like
```go
import (
	"bytes"

	. "github.com/pingcap/check"
	. "github.com/pingcap/tidb-tools/pkg/schemacmp"
)
```

but seems you have updated now.

pkg/schemacmp/table.go
csuzhangxc: Is it simpler to remove `Not`? because of we always use `!index[indexInfoTupleIndexNotPrimary]*`

kennytm: That would make the semantic unclear. Or we could introduce `InvertedBool` where `false > true` 🤔

csuzhangxc (commented):
LGTM

pkg/schemacmp/table.go
```diff
@@ -0,0 +1,402 @@
...
+
+	return Tuple{
+		EqualitySingleton(indexColumns),
+		Bool(!ii.Unique),
+		Bool(!ii.Primary),
+		Singleton(ii.Tp),
+	}
+}
+
+func restoreIndexInfoFromUnwrapped(index []interface{}, keyName string, ctx *format.RestoreCtx) {
```
WangXiangUSTC: how about move ctx to the first position in the arguments

pkg/schemacmp/lattice_test.go
```diff
@@ -0,0 +1,330 @@
...
+			a:            Bool(false),
+			b:            Maybe(Bool(false)),
+			compareError: `type mismatch.*`,
+			joinError:    `type mismatch.*`,
+		},
+		{
+			a:             StringList{"one", "two", "three"},
+			b:             StringList{"one", "two", "three", "four", "five"},
+			compareResult: -1,
+			join:          StringList{"one", "two", "three", "four", "five"},
```
WangXiangUSTC, line 222: do we need add test case 
```
a: StringList{"two", "three"},
b: StringList{"one", "two", "three"},
```
and the compare result should be -1?

pkg/schemacmp/table_test.go
```diff
@@ -0,0 +1,410 @@
...
+		b       string
+		cmp     int
+		cmpErr  string
+		join    string
+		joinErr string
+	}{
+		{
+			name: "DM_002/1",
+			a:    "CREATE TABLE tb1 (col1 INT)",
+			b:    "CREATE TABLE tb2 (col1 INT, new_col1 INT)",
```
WangXiangUSTC, line 74: do we need to add test case like
```
a:    "CREATE TABLE tb1 (col1 INT)",
b:    "CREATE TABLE tb2 (new_col1INT, col1 INT)",
```

WangXiangUSTC (commented):
rest LGTM

pkg/schemacmp/lattice_test.go
kennytm: OK, but the result should be "Incompatible", since a `StringList` only compare the start of the list.

This is because `ENUM('two', 'three')` (two = 0, three = 1) is incompatible with `ENUM('one', 'two', 'three')` (one = 0, two = 1, three = 2), and similarly `SET('two', 'three')` (two = 1, three = 2) is incompatible with `SET('one', 'two', 'three')` (one = 1, two = 2, three = 4).

pkg/schemacmp/table_test.go
kennytm: OK, though since we're using a map which is unordered, it doesn't really matter 😄

WangXiangUSTC (approved):
LGTM

PR #313, opened 2020-01-06
restore-util: add warning about unmatched table id

5kbpers:
Test if the table id in start key and end key is matched.

NA

### Check List 

Tests 

 - Unit test
 - Integration test

Side effects


overvenus (commented):
Could you add some tests to cover `sortRanges`? Backup range is bit different after https://github.com/pingcap/br/pull/72. Instead of `[tableID(1), tableID(1)+1)` it now backups `[7480000000000000015f720000000000000000, 7480000000000000015f72ffffffffffffffff00)`

https://gist.github.com/overvenus/b1135f7b02f46004cf02919b6fea5d87

pkg/restore-util/range_test.go
```diff
@@ -3,52 +3,78 @@ package restore_util
...
+	expected := params[1].([]Range)
+	if len(obtained) != len(expected) {
+		return false, ""
 	}
-	for i, rg := range ranges {
-		res := resRanges[i]
-		if !bytes.Equal(rg.StartKey, res.StartKey) || !bytes.Equal(rg.EndKey, res.EndKey) {
-			t.Fatalf("some inserted ranges missed: rg=%v res=%v", ranges, resRanges)
+	for i := range obtained {
+		if bytes.Compare(obtained[i].StartKey, expected[i].StartKey) != 0 ||
```
overvenus:
```suggestion
		if !bytes.Equal(obtained[i].StartKey, expected[i].StartKey) ||
```

pkg/restore-util/range_test.go
```diff
@@ -3,52 +3,78 @@ package restore_util
...
+			bytes.Compare(obtained[i].EndKey, expected[i].EndKey) != 0 {
```
overvenus:
```suggestion
			!bytes.Equal(obtained[i].EndKey, expected[i].EndKey) {
```

pkg/restore-util/split_test.go
```diff
@@ -184,14 +185,15 @@ func TestSplit(t *testing.T) {
...
 	err := regionSplitter.Split(ctx, ranges, rewriteRules, func(key [][]byte) {})
 	if err != nil {
-		t.Fatalf("split regions failed: %v", err)
+		c.Assert(err, IsNil, Commentf("split regions failed: %v", err))
 	}
 	regions := client.GetAllRegions()
 	if !validateRegions(regions) {
 		for _, region := range regions {
-			t.Errorf("region: %v", region.Region)
+			fmt.Printf("region: %v\n", region.Region)
```
overvenus:
```suggestion
			c.Logf("region: %v\n", region.Region)
```

pkg/restore-util/split_test.go
```diff
@@ -184,14 +185,15 @@ func TestSplit(t *testing.T) {
...
 		}
-		t.Fatalf("get wrong result")
+		fmt.Printf("get wrong result")
```
overvenus:
```suggestion
		c.Log("get wrong result")
```

3pointer (dismissed):
@overvenus  PTAL

overvenus (dismissed):
LGTM

pkg/restore-util/range.go
```diff
@@ -41,25 +42,34 @@ func sortRanges(ranges []Range, rewriteRules *RewriteRules) ([]Range, error) {
...
-			if endRule == nil {
-				log.Warn("cannot find rewrite rule", zap.Binary("key", rg.EndKey))
-			} else {
-				log.Debug(
-					"rewrite end key",
-					zap.Binary("key", rg.EndKey),
-					zap.Stringer("rule", endRule))
+				log.Warn("table id does not match",
+					zap.ByteString("startKey", rg.StartKey),
+					zap.ByteString("endKey", rg.EndKey),
```
kennytm: Use `zap.Binary` for these (also the same for all other `zap.ByteString` introduced in this PR)

overvenus (dismissed):

kennytm:
The behavior of this PR is that, if `startTableID` ≠ `endTableID`, the rewrite rules will all be deactivated. Is this the expected behavior?

BTW TiKV uses the separate "rewrite_prefix" for the start key and "[rewrite_prefix_of_end_key](https://github.com/tikv/tikv/blob/26097303ded323637e3857fe12c871868e37af77/components/keys/src/rewrite.rs#L27-L47)" for the end key to tackle the `[t, t+1)` issue. But in this library we only have a single `replacePrefix` 🤔

5kbpers:
@kennytm We can return an error instead of outputting a warning. 
Since there isn't the `[t, t+1)` issue in BR anymore, I guess we can unify the interface for the start key and the end key.

kennytm:
Let's return an error then.

5kbpers:
@kennytm Having changed that, PTAL, thanks!

kennytm:
```
[2020-01-07T06:50:41.415Z] FAIL: range_test.go:43: testRestoreUtilSuite.TestSortRange
[2020-01-07T06:50:41.415Z] 
[2020-01-07T06:50:41.415Z] range_test.go:65:
[2020-01-07T06:50:41.415Z]     c.Assert(err, IsNil, Commentf("sort range1 failed: %v", err))
[2020-01-07T06:50:41.415Z] ... value *errors.fundamental = table id does not match ("table id does not match")
[2020-01-07T06:50:41.415Z] ... sort range1 failed: table id does not match
```

Rest LGTM

kennytm (approved):

PR #317, opened 2020-02-14
diff: increase the default timeout for db operate

WangXiangUSTC:

users feedback that the 5 second is not enough for database operate some times.

increase to 10 second


niezefeng (approved):
LGTM

csuzhangxc (approved):
LGTM

PR #318, opened 2020-02-18
filter: added options to disable regex and/or wildcard

kennytm:

When the table name to match necessarily contains regex or wildcard characters (`~`, `*`, `?`, `[`), we need to escape the table name. This is quite inelegant.

Add a feature to `filter` to disable regex and/or wildcard support from config, so we don't need to manually escape the input.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

 - Increased code complexity

Related changes

    * All components relying on this component for black-white-list have two extra fields in the config.


kennytm:
Going to handle it through the new filter system.

PR #321, opened 2020-03-05
diff: support auto_random in TiDB

WangXiangUSTC:

diff will get `Invalid auto_random` error when getting table's schema with auto_random

set `AllowAutoRandom = true` for tidb

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -32,6 +32,7 @@ import (
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/tidb-tools/pkg/dbutil"
 	"github.com/pingcap/tidb-tools/pkg/utils"
+	tidbconfig "github.com/pingcap/tidb/config"
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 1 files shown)


july2993:
what code of diff depend on this?

july2993 (commented):
LGTM
I think you should avoid depending on the code to build `model.TableInfo` in tidb from create table SQL.
you can get the table info using sql like `SHOW COLUMNS FROM ..` and `show index from ..`

WangXiangUSTC:
> what code of diff depend on this?

https://github.com/pingcap/tidb-tools/blob/cccf51b6f44cc6bbd96841d2fdc0c7bd9364ac69/pkg/dbutil/table.go#L53
use tidb's function to get `TableInfo`

WangXiangUSTC:
> LGTM
> I think you should avoid depending on the code to build `model.TableInfo` in tidb from create table SQL.
> you can get the table info using sql like `SHOW COLUMNS FROM ..` and `show index from ..`

ok, we can do this later

lichunzhu (approved):
LGTM

PR #316, opened 2020-02-09
filter: use TiDB's IsMemOrSysDB() method in IsSystemSchema()

kennytm:
### What problem does this PR solve? 

TiDB introduced two additional system schemas in pingcap/tidb#13757 (`metrics_schema`) and pingcap/tidb#14147 (`inspection_schema`), which `filter.IsSystemSchema()` currently cannot recognize.

### What is changed and how it works?

To avoid playing catch up with TiDB in the future, we changed the implementation of `filter.IsSystemSchema()` to forward to the standard `"github.com/pingcap/tidb/util".IsMemOrSysDB()`. 

### Check List 

Tests 

 - Unit test

Code changes

 - Has exported function/method change

Side effects

Related changes

PR #322, opened 2020-03-06
hard code to remove tidb dep from pkg/filter

IANTHEREAL:


3pointer (approved):

amyangfei (approved):
LGTM

kennytm:
🤨 The reason we used `util.IsMemOrSysDB` (#316) was to avoid someday TiDB added yet another system schema.

PR #323, opened 2020-03-11
pkg: refine tls && update tidb version

WangXiangUSTC:

1. need to support online reload for tls https://github.com/pingcap/dm/issues/467
2. need to support CN certification https://github.com/pingcap/dm/issues/524

1. refine pkg for tls, for example, provide some function for grpc to use reload tls file, most code is copied from  https://github.com/pingcap/tidb-lightning/blob/master/lightning/common/security.go
2. add certification for Common Name in CA

### Check List 

Tests 

 - Unit test


pkg/utils/security.go
```diff
@@ -17,41 +17,170 @@ import (
...
-			// Load the client certificates from disk
-			certificate, err := tls.LoadX509KeyPair(SSLCert, SSLKey)
-			if err != nil {
-				return nil, errors.Errorf("could not load client key pair: %s", err)
-			}
-			certificates = append(certificates, certificate)
-		}
+func ToTLSConfig(caPath, certPath, keyPath string) (*tls.Config, error) {
+	return ToTLSConfig(caPath, certPath, keyPath)
+}
```
kennytm: ????????

WangXiangUSTC: a mistake, should be `ToTLSConfigWithVerify(caPath, certPath, keyPath, nil)`

WangXiangUSTC:
@kennytm @csuzhangxc PTAL

pkg/utils/security.go
```diff
@@ -17,41 +17,170 @@ import (
...
+		ClientCAs:    certPool,
+		NextProtos:   []string{"h2", "http/1.1"}, // specify `h2` to let Go use HTTP/2.
+	}
+
+	checkCN := make(map[string]struct{})
+	for _, cn := range verifyCN {
+		cn = strings.TrimSpace(cn)
+		checkCN[cn] = struct{}{}
+	}
+	if len(checkCN) != 0 {
```
kennytm: move lines 80-84 inside the `if` branch.

```suggestion
	if len(verifyCN) != 0 {
		checkCN := make(map[string]struct{})
		for _, cn := range verifyCN {
			cn = strings.TrimSpace(cn)
			checkCN[cn] = struct{}{}
		}
```

pkg/utils/security.go
```diff
@@ -17,41 +17,170 @@ import (
...
+		tlsCfg.ClientAuth = tls.RequireAndVerifyClientCert
+
+		tlsCfg.VerifyPeerCertificate = func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
+			if len(checkCN) == 0 {
+				return nil
+			}
+
```
kennytm:
```suggestion
```

pkg/utils/security.go
```diff
@@ -17,41 +17,170 @@ import (
...
+	client := ClientWithTLS(inner)
+
+	return &TLS{
+		inner:  inner,
+		client: client,
+		url:    "https://" + host,
+	}, nil
+}
+
+// ClientWithTLS ...
```
kennytm: fill in the `...` 🙂

pkg/utils/tls_test/ca.csr
```diff
@@ -0,0 +1,7 @@
+-----BEGIN CERTIFICATE REQUEST-----
```
kennytm: no need to include the CSR files.

but perhaps add the instruction how to regenerate the *.key and *.pem

pkg/utils/security_test.go
```diff
@@ -0,0 +1,162 @@
...
+	resp, err := ClientWithTLS(clientTLS1).Get(url)
+	c.Assert(err, IsNil)
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	c.Assert(err, IsNil)
+	c.Assert(string(body), Equals, "This an example server")
+
+	// client2 can't visit server
+	_, err = ClientWithTLS(clientTLS2).Get(url)
+	c.Assert(err, ErrorMatches, ".*tls: bad certificate")
```
kennytm, line 138: will the server accept `client2` if the common name is `localhost` (the CA's common name)?

pkg/utils/security.go
WangXiangUSTC: refine in dea0b3a

pkg/utils/security.go
WangXiangUSTC: removed in dea0b3a

pkg/utils/security.go
WangXiangUSTC: fill in dea0b3a

pkg/utils/tls_test/ca.csr
WangXiangUSTC: do you mean:
1. delete the CSR files
2. add some comments in *.key and *.pem to introduce how to generate these files

pkg/utils/security_test.go
WangXiangUSTC: no, you will see the log of unit test:
```
2020/03/12 12:21:44 http: TLS handshake error from 127.0.0.1:51864: client certificate authentication failed. The Common Name from the client certificate [client2 localhost] was not found in the configuration cluster-verify-cn with value: [client1]
```

the common name already include `localhost`

pkg/utils/tls_test/ca.csr
kennytm: 1. delete irrelevant files like the CSR and SRL files.
2. add a readme or a *.sh file about how we can regenerate the pem & key for the test.

WangXiangUSTC: ok, I do it in b91ec85

kennytm (commented):
LGTM

july2993 (approved):

PR #325, opened 2020-03-17
schemacmp: properly support PKIsHandle columns

kennytm:

Previously, an `INT PRIMARY KEY` column was not counted as index, causing the join to lose its "Primary key" flag, and making it not the tightest bound.

These PKIsHandle columns are now included into the index map in the encoded representation.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


csuzhangxc (commented):
LGTM

WangXiangUSTC (approved):
LGTM

PR #330, opened 2020-03-25
sync-diff: add option to check views

lichunzhu:
`sync-diff` will filter views now and don't check them.

Add an option `ignore-view-check` in config. The default value of `ignore-view-check` is true. When `ignore-view-check` is false sync-diff will fetch views of databases and check the difference.

### Check List 

Tests 

 - Unit test
 - Integration test


lichunzhu:
@WangXiangUSTC @csuzhangxc PTAL

csuzhangxc:
I prefer to have `do-view-check` with default `false`.

lichunzhu:
It can't work with simple development. Close this PR for now.

PR #331, opened 2020-04-09
diff: escape column name

WangXiangUSTC:

if the column name is keyword in mysql/tidb, will return error when select data.

escape column name

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -821,7 +821,7 @@ func getChunkRows(ctx context.Context, db *sql.DB, schema, table string, tableIn
 
 	columnNames := make([]string, 0, len(tableInfo.Columns))
 	for _, col := range tableInfo.Columns {
-		columnNames = append(columnNames, col.Name.O)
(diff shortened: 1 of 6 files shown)


WangXiangUSTC:
@lichunzhu @csuzhangxc PTAL

lichunzhu (commented):
LGTM

pkg/diff/diff.go
```diff
@@ -829,6 +829,10 @@ func getChunkRows(ctx context.Context, db *sql.DB, schema, table string, tableIn
 		collation = fmt.Sprintf(" COLLATE \"%s\"", collation)
 	}
 
+	for i, key := range orderKeys {
+		orderKeys[i] = dbutil.ColumnName(key)
```
csuzhangxc, line 833: Do we need to convert for `schema`, `table` and `columns` bellow?

csuzhangxc (commented):
rest LGTM

csuzhangxc (dismissed):
LGTM

csuzhangxc (approved):

PR #333, opened 2020-04-15
pumpclient: remove dependency on the tidb repo

tiancaiamao:

No circle dependency so that we can upgrade to module/v4

TiDB already Init the log, so provide the `InitLogger` function is unnecessary.

### Check List 

Tests 

 - No code

Code changes

Side effects


tiancaiamao:
PTAL @WangXiangUSTC @lysu

lysu (approved):
LGTM

WangXiangUSTC:
if remove the initial log, the pump client can not print message in tidb's log, should we add it later?

tiancaiamao:
We're not adopting module/v4, so this change is useless.

PR #342, opened 2020-05-14
filter/: fix black white list caseSensitive bug

lichunzhu:
When we deploy ApplyOn/Match method of black-white list filter on tables we will change values of original tables.

Use new cloned copy to do filter work.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


3pointer (approved):
LGTM

PR #345, opened 2020-05-18
diff: fix panic when get empty random value for split chunk

WangXiangUSTC:

when executing `splitRangeByRandom`, if `randomValues` is empty will panic at https://github.com/pingcap/tidb-tools/blob/330c1a4a8ec0692d83d0f91fdf2ad9d297475faf/pkg/diff/chunk.go#L268

add a judgement

### Check List 

Tests 

 - Unit test


zier-one (approved):
LGTM

csuzhangxc (approved):
LGTM

PR #347, opened 2020-05-27
table-filter: added filter.All() to conveniently make a match-all filter

kennytm:

Add a function `filter.All()` which creates a match-all filter. This is used as the default value of `filter.Filter` in a config, which has the same effect as `f, _ := filter.Parse([]string{"*.*"})`.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


lichunzhu (approved):
LGTM

july2993 (approved):

PR #348, opened 2020-06-04
make ParseMySQLReplicationRules able to parser nil rules

lichunzhu:
fix ParseMySQLReplicationRules nil pointer.

return `All()` filter when rules is nil.

### Check List 

Tests 

 - Unit test
 - Integration test


csuzhangxc (commented):
LGTM

kennytm (approved):
LGTM

PR #351, opened 2020-06-08
Add test for 'exclude-tables' config

qixiaobin:
Add test for 'exclude-test'

1.Create a table and write it name into config_base.toml to exclude this table. 
2.Run sync_diff_inspector
3.This table name should not appear in the output file

### Check List 

Tests 
 - Integration test

Code changes

 - Just change run.sh and the config file for test.

Side effects

 - No

Related changes

 - No


WangXiangUSTC (commented):
thanks, LGTM

csuzhangxc (approved):
LGTM

PR #352, opened 2020-06-12
diff: fix block

WangXiangUSTC:

when `ctx.Done()`, will exit `UpdateSummaryInfo`, and `stopUpdateSummaryCh <- true` will be blocked because of no one read data from this channel.

only when ctx is not done send true to `stopUpdateSummaryCh`


csuzhangxc (approved):
LGTM

lichunzhu (approved):
LGTM

Issue #354, opened 2020-06-16
avoid save table sumamary info failed warn log at start up


july2993:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.

2. What did you expect to see?
no warn log

3. What did you see instead?
```
[2020/06/16 04:22:34.399 +00:00] [WARN] [diff.go:654] ["save table summary info failed"]
...
```

4. What version of TiDB are you using (`tidb-server -V` or run `select tidb_version();` on TiDB)?

5. which tool are you using?
sync-diff
6. what versionof tool are you using (`pump -V` or `tidb-lightning -V` or `syncer -V`)?

july2993:
cc @WangXiangUSTC

WangXiangUSTC:
fix in https://github.com/pingcap/tidb-tools/pull/355

WangXiangUSTC:
reason：
1. diff will update and print summary info every 10 seconds, and the summary info is calculated by select chunk table.
2. diff will save chunk's information to chunk table after all chunk is split.

when update summary info, the table may not finish split chunk, so there is no data in chunk table, and diff will print warn log, users may be confused by this.

PR #355, opened 2020-06-16
diff: some fix

WangXiangUSTC:

1. fix issue #354 
  - diff will update and print summary info every 10 seconds, and the summary info is calculated by select `chunk` table.
  - diff will save chunk's information to `chunk` table after all chunk is split.  

  when update summary info, the table may not finish split chunk, so there is no data in `chunk` table, and diff will print warn log, users may be confused by this.

2. update parser and tidb by the way, and a minor fix on set `AllowAutoRandom` to support parser `auto random`

3. support table name and column name with ``` 

1.  just print a debug log and don't return error for `getChunkSummary`
2. update parser and tidb version
3.  replace ``column name`` with `dbutil.ColumnName(column name)`


lichunzhu (dismissed):
LGTM

july2993 (dismissed):

pkg/diff/diff_test.go
```diff
@@ -48,50 +48,52 @@ func (*testDiffSuite) TestGenerateSQLs(c *C) {
...
-	c.Assert(replaceSQL, Equals, "REPLACE INTO `test`.`atest`(`id`,`name`,`birthday`,`update_time`,`money`) VALUES (NULL,'a\\'a','2018-01-01 00:00:00','10:10:10',11.1111);")
-	c.Assert(deleteSQL, Equals, "DELETE FROM `test`.`atest` WHERE `id` is NULL AND `name` = 'a\\'a' AND `birthday` = '2018-01-01 00:00:00' AND `update_time` = '10:10:10' AND `money` = 11.1111;")
+	replaceSQL = generateDML("replace", rowsData, tableInfo, "diff_test")
+	deleteSQL = generateDML("delete", rowsData, tableInfo, "diff_test")
+	c.Assert(replaceSQL, Equals, "REPLACE INTO `diff_test`.`atest`(`id`,`name`,`birthday`,`update_time`,`money`) VALUES (NULL,'a\\'a','2018-01-01 00:00:00','10:10:10',11.1111);")
+	c.Assert(deleteSQL, Equals, "DELETE FROM `diff_test`.`atest` WHERE `id` is NULL AND `name` = 'a\\'a' AND `birthday` = '2018-01-01 00:00:00' AND `update_time` = '10:10:10' AND `money` = 11.1111;")
 }
 
-func (t *testDiffSuite) testDiff(c *C) {
+func (t *testDiffSuite) TestDiff(c *C) {
```
WangXiangUSTC, line 86: this function is not run before 🤢

july2993: Screen Shot 2020-06-18 at 3 31 50 PM

suggest use golangci-lint to help improve the code

WangXiangUSTC: 👍

WangXiangUSTC: I will update this in next pr

pkg/diff/diff.go
```diff
@@ -90,7 +89,7 @@ type TableDiff struct {
 	IgnoreDataCheck bool `json:"-"`
 
 	// set true will continue check from the latest checkpoint
-	UseCheckpoint bool `json:"use-checkpoint"`
+	UseCheckpoint bool `json:"-"`
```
WangXiangUSTC, line 92: fix issue https://github.com/pingcap/tidb-tools/issues/358

lichunzhu (approved):
LGTM

Issue #360, opened 2020-06-18
sync-diff should log config at start up


july2993:
## Feature Request

**Is your feature request related to a problem? Please describe:**

sync-diff can log config at start up.

so it can easy to check the config from the log

**Describe the feature you'd like:**

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

WangXiangUSTC:
fixed in https://github.com/pingcap/tidb-tools/pull/364

Issue #361, opened 2020-06-18
avoid the slow query warn log in sycn-diff


july2993:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.

many
```
[2020/06/18 03:49:33.670 +00:00] [WARN] [common.go:651] ["exec sql slow"] [sql="REPLACE INTO `sync_diff_inspector`.`chunk` VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?);"]......
...
..
```
2. What did you expect to see?
no such many annoying warn log, and this is fine actually

3. What did you see instead?

4. What version of TiDB are you using (`tidb-server -V` or run `select tidb_version();` on TiDB)?

5. which tool are you using?
sync-diff
6. what versionof tool are you using (`pump -V` or `tidb-lightning -V` or `syncer -V`)?

WangXiangUSTC:
fixed in #364

PR #364, opened 2020-06-19
diff: refine log

WangXiangUSTC:

fix issue https://github.com/pingcap/tidb-tools/issues/360 https://github.com/pingcap/tidb-tools/issues/361

1. print basic information at the beginning, include git hash, branch and so on
2. print config before check
3. print debug log for slow SQL, and increase the threshold value for slow query.

### Check List 

Tests 

 - Manual test (run diff and check the log)


pkg/dbutil/common.go
```diff
@@ -648,14 +650,14 @@ func ExecSQLWithRetry(ctx context.Context, db *sql.DB, sql string, args ...inter
 		_, err = db.ExecContext(ctx, sql, args...)
 		takeDuration := time.Since(startTime)
 		if takeDuration > SlowWarnLog {
-			log.Warn("exec sql slow", zap.String("sql", sql), zap.Reflect("args", args), zap.Duration("take", takeDuration))
+			log.Debug("exec sql slow", zap.String("sql", sql), zap.Reflect("args", args), zap.Duration("take", takeDuration))
```
csuzhangxc, line 653: No `Warn` for `SlowWarnLog` now, how about giving `SlowWarnLog` another name?

WangXiangUSTC: update in 758aca3

csuzhangxc (commented):
LGTM

PR #365, opened 2020-06-22
check: remove restrictions for MySQL 8.0

csuzhangxc:

starting to support MySQL 8.0 in DM.

remove restrictions for MySQL 8.0 in version checker.

### Check List 

Tests 

 - Unit test

Code changes


WangXiangUSTC (commented):
LGTM

PR #366, opened 2020-06-24
diff: refine log && minor fix

WangXiangUSTC:
### What problem does this PR solve? 

1. fix issue https://github.com/pingcap/tidb-tools/issues/357, don't print this log
2. refine log for check report, the old log like:

`[2020/06/24 16:36:23.332 +08:00] [INFO] [main.go:87] ["check report"] [report="\ncheck result: fail!\n0 tables' check passed, 1 tables' check failed.\n\ntable: diff_test.t_10\ntable's struct equal\ntable's data not equal\n\n"]`

after this PR, it will be:

`[2020/06/24 17:50:47.149 +08:00] [INFO] [report.go:80] ["check result summary"] ["check passed num"=0] ["check failed num"=1]
[2020/06/24 17:50:47.149 +08:00] [INFO] [report.go:87] ["table check result"] [schema=diff_test] [table=t_10] ["struct equal"=true] ["data equal"=false]`

3. a minor fix, after this pr https://github.com/pingcap/tidb-tools/pull/346/files#r444783910, if meet error will continue and doesn't save the check result. 

### Check List 

Tests 

 - Manual test (check log)

PR #368, opened 2020-06-29
diff: minor fix on integration test

WangXiangUSTC:

#366 change the log, but not update the integration test.

update integration test for check log

### Check List 

Tests 

 - Integration test


july2993 (approved):

GMHDBJD (approved):
LGTM

Issue #362, opened 2020-06-18
slow to init chunks and some unexpect summary info log


july2993:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.

grep the origin log by `grep "summary info"`
[su.txt](https://github.com/pingcap/tidb-tools/files/4798210/su.txt)

it's slow to init chunks of the table `txn_history` and inserting  to the chunk table
there's no easy way to check the time.
```
[2020/06/18 02:33:33.450 +00:00] [INFO] [checkpoint.go:231] ["summary info"] [instance_id=target] [schema=db] [table=offset] ["chunk num"=1] ["success num"=1] ["failed num"=0] ["ignore num"=0]
[2020/06/18 03:41:53.686 +00:00] [INFO] [checkpoint.go:231] ["summary info"] [instance_id=target] [schema=db] [table=txn_history] ["chunk num"=631] ["success num"=0] ["failed num"=0] ["ignore num"=0]
[2020/06/18 03:42:03.712 +00:00] [INFO] [checkpoint.go:231] ["summary info"] [instance_id=target] [schema=db] [table=txn_history] ["chunk num"=1321] ["success num"=0] ["failed num"=0] ["ignore num"=0]
[2020/06/18 03:42:13.690 +00:00] [INFO] [checkpoint.go:231] ["summary info"] [instance_id=target] [schema=db] [table=txn_history] ["chunk num"=2001] ["success num"=0] ["failed num"=0] ["ignore num"=0]
[2020/06/18 03:42:23.681 +00:00] [INFO] [checkpoint.go:231] ["summary info"] [instance_id=target] [schema=db] [table=txn_history] ["chunk num"=2674] ["success num"=0] ["failed num"=0] ["ignore num"=0]
[2020/06/18 03:42:33.691 +00:00] [INFO] [checkpoint.go:231] ["summary info"] [instance_id=target] [schema=db] [table=txn_history] ["chunk num"=3370] ["success num"=0] ["failed num"=0] ["ignore num"=0]
[2020/06/18 03:42:43.692 +00:00] [INFO] [checkpoint.go:231] ["summary info"] [instance_id=target] [schema=db] [table=txn_history] ["chunk num"=4073] ["success num"=0] ["failed num"=0] ["ignore num"=0]
[...]

PR #369, opened 2020-06-30
diff: save summary info in memory

WangXiangUSTC:

fix issue https://github.com/pingcap/tidb-tools/issues/362
getting the wrong summary like 
```
[2020/06/18 07:46:23.864 +00:00] [INFO] [checkpoint.go:231] ["summary info"] [instance_id=target] [schema=db] [table=txn_history] ["chunk num"=247656] ["success num"=153326] ["failed num"=0] ["ignore num"=0]
[2020/06/18 07:46:33.810 +00:00] [INFO] [checkpoint.go:231] ["summary info"] [instance_id=target] [schema=db] [table=txn_history] ["chunk num"=247656] ["success num"=153496] ["failed num"=0] ["ignore num"=0]
[2020/06/18 07:46:43.873 +00:00] [INFO] [checkpoint.go:231] ["summary info"] [instance_id=target] [schema=db] [table=txn_history] ["chunk num"=247656] ["success num"=0] ["failed num"=0] ["ignore num"=0]
```
already get 153496 success num, but get 0 success num later, don't know why now, but we can save the summary information in memory instead of select the `chunk` table every time print summary information to fix this issue,  and this will save TiDB/MySQL's performance and diff can run faster.

save summary info in memory

### Check List 

Tests 

 - Unit test

### DIFF
--- a/pkg/diff/checkpoint.go
+++ b/pkg/diff/checkpoint.go
@@ -18,6 +18,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/pingcap/errors"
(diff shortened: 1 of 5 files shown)


pkg/diff/checkpoint.go
```diff
@@ -223,21 +222,24 @@ func initTableSummary(ctx context.Context, db *sql.DB, schema, table string, con
...
 	if total == 0 {
 		// don't need to update summary info
 		return nil
 	}
 
 	log.Info("summary info", zap.String("instance_id", instanceID), zap.String("schema", schema), zap.String("table", table), zap.Int64("chunk num", total), zap.Int64("success num", successNum), zap.Int64("failed num", failedNum), zap.Int64("ignore num", ignoreNum))
 
-	state := notCheckedState
-	if total == successNum+failedNum+ignoreNum {
+	checkedNum := successNum + failedNum + ignoreNum
```
WangXiangUSTC, line 238: minor fix

WangXiangUSTC:
@GMHDBJD @july2993 PTAL

GMHDBJD (commented):
LGTM

pkg/diff/diff.go
```diff
@@ -450,21 +462,26 @@ func (t *TableDiff) checkChunkDataEqual(ctx context.Context, filterByRand bool,
 
 func (t *TableDiff) compareChecksum(ctx context.Context, chunk *ChunkRange) (bool, error) {
```
july2993, line 463: we can get the source and target checksum concurrent?

july2993 (commented):
I think one reason it's slow is caused by it insert the chunk one by one, it's good to just saved in memory, but i don't think it can run faster for avoiding reading it per 10 second.
LGTM but I don't think it fixes the issue

WangXiangUSTC:
> I think one reason it's slow is caused by it insert the chunk one by one, it's good to just saved in memory, but i don't think it can run faster for avoiding reading it per 10 second.
> LGTM but I don't think it fixes the issue

I do these refine in this PR https://github.com/pingcap/tidb-tools/pull/370, include insert chunks information in batch SQLs, and calculate source and target's checksum concurrent @july2993

pkg/diff/diff.go
WangXiangUSTC: do it in #370

july2993 (approved):

Issue #356, opened 2020-06-18
sync-diff-inspector: avoid eating too many memory


july2993:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.

It's easy to split a huge chunk.
having already 380G memory still oom, this is unacceptable.
```
fatal error: runtime: out of memory

runtime stack:
runtime.throw(0x1915ed4, 0x16)
        /usr/local/go/src/runtime/panic.go:774 +0x72
runtime.sysMap(0x117c4000000, 0x4000000, 0x2e7acd8)
        /usr/local/go/src/runtime/mem_linux.go:169 +0xc5
runtime.(*mheap).sysAlloc(0x2e60f20, 0x4000, 0x41f2a2, 0x7ff63cc03200)
```

2. What did you expect to see?
- better split chunk
- do not oom even if a chunk contains too many rows.
   - we can fetch data by `order by ... limit ...` can compare step by step(we can split chunk dynamic doing this?).

3. What did you see instead?

4. What version of TiDB are you using (`tidb-server -V` or run `select tidb_version();` on TiDB)?

5. which tool are you using?
sync-diff
6. what versionof tool are you using (`pump -V` or `tidb-lightning -V` or `syncer -V`)?

july2993:
cc @WangXiangUSTC

kennytm:
@july2993 please specify which tool you are talking about

july2993:
> @july2993 please specify which tool you are talking about

sync-diff, updated in the description.

PR #370, opened 2020-07-02
diff: refine performance and memory usage

WangXiangUSTC:

1. Refine the performance
2. Optimize memory usage https://github.com/pingcap/tidb-tools/issues/356

1. initial chunks information to the checkpoint table with batch insert SQL, it will save much time if have too many chunks.
2. calculate checksum in source and target database at the same time by use goroutine.
3. read data from the database only when needed, instead of select all the chunk's data, this will save memory usage.

### Check List 

Tests 

 - Unit test
 - Integration test

### DIFF
--- a/pkg/diff/checkpoint.go
+++ b/pkg/diff/checkpoint.go
@@ -18,6 +18,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"sync"
 	"time"
 
@@ -113,6 +114,45 @@ func saveChunk(ctx context.Context, db *sql.DB, chunkID int, instanceID, schema,
 		log.Error("save chunk info failed", zap.Error(err))
 		return errors.Trace(err)
 	}
+	return nil
+}
+
+// initChunks initials the chunks' information into chunk table
+func initChunks(ctx context.Context, db *sql.DB, instanceID, schema, table string, chunks []*ChunkRange) error {
+	beginTime := time.Now()
+	batch := 100
+	num := 0
+	sqlPrefix := fmt.Sprintf("INSERT INTO `%s`.`%s` VALUES", checkpointSchemaName, chunkTableName)
+
+	valuesPlaceholders := "(?, ?, ?, ?, ?, ?, ?, ?, ?)"
+	valuesPlaceholdersArray := make([]string, 0, batch)
+	values := make([]interface{}, 0, 9*batch)
+
+	for i, chunk := range chunks {
+		num++
+		chunkBytes, err := json.Marshal(chunk)
+		if err != nil {
+			return errors.Trace(err)
+		}
+
+		values = append(values, chunk.ID, instanceID, schema, table, chunk.Where, "", string(chunkBytes), chunk.State, time.Now())
+		valuesPlaceholdersArray = append(valuesPlaceholdersArray, valuesPlaceholders)
+
+		if num >= batch || i == len(chunks)-1 {
+			sql := fmt.Sprintf("%s%s", sqlPrefix, strings.Join(valuesPlaceholdersArray, ", "))
+			err = dbutil.ExecSQLWithRetry(ctx, db, sql, values...)
+			if err != nil {
+				log.Error("save chunk info failed", zap.Error(err))
+				return errors.Trace(err)
+			}
+			num = 0
+			valuesPlaceholdersArray = make([]string, 0, batch)
+			values = make([]interface{}, 0, 9*batch)
+		}
+	}
+
+	log.Info("initial chunks", zap.Duration("cost", time.Since(beginTime)))
+
 	return nil
 }
 
--- a/pkg/diff/checkpoint.go
+++ b/pkg/diff/checkpoint.go
@@ -18,6 +18,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/pingcap/errors"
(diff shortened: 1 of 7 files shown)


WangXiangUSTC:
@july2993 @GMHDBJD PTAL

pkg/diff/diff.go
```diff
@@ -487,140 +527,187 @@ func (t *TableDiff) compareChecksum(ctx context.Context, chunk *ChunkRange) (boo
...
-			})
-			sourceMap[source] = struct{}{}
-			sourceOffset[source]++
+		for _, i := range needDeleteSource {
+			delete(sourceMap, i)
 		}
 
-		if rowDatas.Len() == 0 {
-			break
+		// all the sources had read to the end, not data to return
```
GMHDBJD: Should be no data to return.

sync_diff_inspector/main.go
```diff
@@ -63,7 +63,7 @@ func main() {
 
 	ctx := context.Background()
 	if !checkSyncState(ctx, cfg) {
-		log.Error("check failed!!!")
+		log.Warn("check failed!!!")
```
GMHDBJD, line 66: Why change it to Warn?

GMHDBJD (commented):
rest LGTM

pkg/diff/diff.go
WangXiangUSTC: updated

sync_diff_inspector/main.go
WangXiangUSTC: because `Error` will print the stack, looks ugly

GMHDBJD (commented):
LGTM

pkg/diff/checkpoint.go
```diff
@@ -116,6 +117,45 @@ func saveChunk(ctx context.Context, db *sql.DB, chunkID int, instanceID, schema,
...
+		if num >= batch || i == len(chunks)-1 {
+			sql := fmt.Sprintf("%s%s", sqlPrefix, strings.Join(valuesPlaceholdersArray, ", "))
+			err = dbutil.ExecSQLWithRetry(ctx, db, sql, values...)
+			if err != nil {
+				log.Error("save chunk info failed", zap.Error(err))
+				return errors.Trace(err)
+			}
+			num = 0
+			valuesPlaceholdersArray = make([]string, 0, batch)
+			values = make([]interface{}, 0, 9*batch)
```
csuzhangxc: Can we reuse the slice allocated in L128 ~ L129?

pkg/diff/diff.go
```diff
@@ -460,21 +447,75 @@ func (t *TableDiff) checkChunkDataEqual(ctx context.Context, filterByRand bool,
...
+	args := utils.StringsToInterfaces(chunk.Args)
+	for _, sourceTable := range t.SourceTables {
+		go getChecksum(sourceTable.Conn, sourceTable.Schema, sourceTable.Table, chunk.Where, t.TargetTable.info, args, "source")
+	}
+
+	go getChecksum(t.TargetTable.Conn, t.TargetTable.Schema, t.TargetTable.Table, chunk.Where, t.TargetTable.info, args, "target")
+
+	for i := 0; i < len(t.SourceTables)+1; i++ {
+		select {
+		case checksumInfo := <-checksumInfoCh:
```
csuzhangxc: only one `case` for `select`, can we use `checksumInfo := <-checksumInfoCh` directly?

pkg/diff/diff.go
```diff
@@ -460,21 +447,75 @@ func (t *TableDiff) checkChunkDataEqual(ctx context.Context, filterByRand bool,
...
+			if checksumInfo.err != nil {
+				// only need to return the first error, others are context cancel error
+				if firstErr == nil {
+					firstErr = checksumInfo.err
+					cancel1()
+				}
+
+				continue
```
csuzhangxc: Can we `return` here?

pkg/diff/diff.go
```diff
@@ -487,140 +528,187 @@ func (t *TableDiff) compareChecksum(ctx context.Context, chunk *ChunkRange) (boo
...
-		// judge rows have all order keys to avoid panic
-		if len(rows) > 0 {
-			if !rowContainsCols(rows[0], orderKeyCols) {
-				return false, errors.Errorf("%s.%s.%s's data don't contain all keys %v", sourceTable.InstanceID, sourceTable.Schema, sourceTable.Table, orderKeyCols)
-			}
-		}
-
-		sourceRows[fmt.Sprintf("source-%d", i)] = rows
+		sourceRows[i] = rows
+		sourceMap[i] = false
```
csuzhangxc: Can you give it a meaningful name now?

pkg/diff/diff.go
```diff
@@ -487,140 +528,187 @@ func (t *TableDiff) compareChecksum(ctx context.Context, chunk *ChunkRange) (boo
...
 			t.sqlCh <- sql
-			index1++
-			index2++
+			lastSourceData = nil
+			lastTargetData = nil
 		}
 	}
 
+	if equal {
+		log.Info("rows is equal", zap.String("table", dbutil.TableName(t.TargetTable.Schema, t.TargetTable.Table)), zap.String("where", dbutil.ReplacePlaceholder(chunk.Where, chunk.Args)), zap.Duration("cost", time.Since(beginTime)))
```
csuzhangxc, line 734: Does this output too many log for a huge table?

pkg/diff/checkpoint.go
WangXiangUSTC: update in https://github.com/pingcap/tidb-tools/pull/370/commits/33cdde062dfcf499e2cc4e2a48abcf4030123832

pkg/diff/diff.go
WangXiangUSTC: update in https://github.com/pingcap/tidb-tools/pull/370/commits/33cdde062dfcf499e2cc4e2a48abcf4030123832

pkg/diff/diff.go
WangXiangUSTC: use `continue` so we can get all results from every goroutine, this will be very fast because other goroutines will exit with `context canceled` error

pkg/diff/diff.go
WangXiangUSTC: rename to `sourceHaveData` in https://github.com/pingcap/tidb-tools/pull/370/commits/33cdde062dfcf499e2cc4e2a48abcf4030123832

pkg/diff/diff.go
WangXiangUSTC: will only print for each chunk, I think this log is necessary, users can know what diff is doing now, and also know the time cost.

csuzhangxc (approved):
LGTM

PR #373, opened 2020-07-09
sync_diff_inspector: fix the snapshot using the DateTime string

SE-Bin:
Tiny fix for the Snapshot using the DateTime string like "2020-01-01 20:20:20"
image

Enclose **tidb_snapshot** in Double quotes `""` for adding it in the dbDSN

### Check List 

Tests 

 - Manual test 

Related changes

### DIFF
--- a/sync_diff_inspector/config.go
+++ b/sync_diff_inspector/config.go
@@ -17,6 +17,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"flag"
+	"strconv"
 
 	"github.com/BurntSushi/toml"
 	"github.com/pingcap/errors"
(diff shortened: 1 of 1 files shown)


WangXiangUSTC:
LGTM, and it is better to modify the integration test https://github.com/pingcap/tidb-tools/blob/master/tests/sync_diff_inspector/snapshot/run.sh#L12, we can do it later

WangXiangUSTC:
@GMHDBJD PTAL

GMHDBJD (approved):
LGTM

WangXiangUSTC (approved):

PR #280, opened 2019-11-04
diff: add sql-mode config

WangXiangUSTC:
### What problem does this PR solve? 
some database doesn't use default SQL mode, and the diff will parse failed when getting table struct. 

### What is changed and how it works?
add sql-mode config for database config

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Related changes

 - Need to update the documentation

PR #381, opened 2020-08-12
check, diff: auto discover ansi-quotes

lance6716:
auto discover ansi-quotes so don't need to set it in config, part of https://github.com/pingcap/dm/issues/878

for processing `db` (new connection) related query results, get ansi-quotes from global variable for this db.
(may need to get session variable for connection related query results, not found yet)

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


pkg/utils/util.go
```diff
@@ -56,3 +59,57 @@ func GetJSON(client *http.Client, url string, v interface{}) error {
 
 	return errors.Trace(json.NewDecoder(resp.Body).Decode(v))
 }
+
+// GetGlobalVariable gets server's global variable
```
lance6716: these functions copied from DM https://github.com/pingcap/dm/blob/56eed525e047f1ecf1ce03709eb5ccb1e7b50caf/pkg/utils/db.go#L356

lance6716:
PTAL @csuzhangxc @GMHDBJD

pkg/dbutil/common.go
```diff
@@ -72,8 +72,6 @@ type DBConfig struct {
 	Schema string `toml:"schema" json:"schema"`
 
 	Snapshot string `toml:"snapshot" json:"snapshot"`
-
-	SQLMode string `toml:"sql-mode" json:"sql-mode"`
```
csuzhangxc, line 76: Do other places still need this (`sql_mode` may including others rather than `ANSI_QUOTES`)?

But we may remove it now and add it back if needed later.

pkg/utils/util.go
csuzhangxc: How about putting them into _dbutil_ rather than _utils_?

pkg/dbutil/common.go
lance6716: I guess other places don't require this `SQLMode` otherwise it won't compile. and in https://github.com/pingcap/tidb-tools/pull/280 this field is used for parser only

pkg/utils/util.go
lance6716: done in https://github.com/pingcap/tidb-tools/pull/381/commits/efda573e693caa32aea1e3dfd43e1e70725bf750

lance6716:
very strange that integration test could pass on my Ubuntu 🤔

lance6716:
I could confirm integration test can pass on my Ubuntu
image

and seems CI fails on `Error 9002: TiKV server timeout`

PTAL @csuzhangxc @GMHDBJD

lance6716:
mysteriously CI pass, PTAL @csuzhangxc @GMHDBJD

pkg/dbutil/common.go
```diff
@@ -606,6 +604,60 @@ func GetDBVersion(ctx context.Context, db *sql.DB) (string, error) {
...
+	if rows.Err() != nil {
+		return "", errors.Trace(err)
+	}
+
+	return value, nil
+}
+
+// GetSQLMode returns sql_mode.
+func GetSQLMode(db *sql.DB) (tmysql.SQLMode, error) {
+	sqlMode, err := GetGlobalVariable(db, "sql_mode")
```
GMHDBJD: Should be session variable because now we support session config for db？

pkg/check/table_structure.go
```diff
@@ -161,8 +159,18 @@ func (c *TablesChecker) Name() string {
...
+			},
+		}
+	}
+
 	sqlMode := ""
-	if c.enableANSIQuotes {
+	if ansiQuotes {
 		sqlMode = "ANSI_QUOTES"
 	}
 	parser2, err := dbutil.GetParser(sqlMode)
```
GMHDBJD: How about combine them into one function?

lance6716: done in https://github.com/pingcap/tidb-tools/pull/381/commits/3dccc006997826ff0344dc233afda8776d223410

pkg/dbutil/common.go
lance6716: done in https://github.com/pingcap/tidb-tools/pull/381/commits/3dccc006997826ff0344dc233afda8776d223410

GMHDBJD (commented):
LGTM

pkg/dbutil/common.go
```diff
@@ -780,3 +833,16 @@ func GetParser(sqlModeStr string) (*parser.Parser, error) {
...
+func GetParserForDB(db *sql.DB) (*parser.Parser, error) {
+	ansiQuotes, err := HasAnsiQuotesMode(db)
+	if err != nil {
+		return nil, err
+	}
+	sqlMode := ""
+	if ansiQuotes {
+		sqlMode = "ANSI_QUOTES"
+	}
+	return getParser(sqlMode)
```
csuzhangxc: Can we set `sql_mode` for all modes from `GetSQLMode` rather than only `ANSI_QUOTES`?

csuzhangxc (commented):
rest LGTM

pkg/dbutil/common.go
lance6716: https://github.com/pingcap/tidb-tools/pull/381/commits/f70d144ff85a84a3b31bb29cdf34db015327b069

Okay, going to align DM's parser to this behaviour

csuzhangxc (approved):
LGTM

PR #382, opened 2020-08-17
table-filter: fix the bug about the table filter parser

zier-one:
 - fix the parser ignoring `!` and `^` in some cases
 - remove the debug code


kennytm (approved):
LGTM

PR #383, opened 2020-08-19
check: support LOAD FROM S3, SELECT INTO S3 for Aurora

lance6716:
support LOAD FROM S3, SELECT INTO S3 to avoid failing parser, https://github.com/pingcap/dm/issues/900

simply remove LOAD FROM S3, SELECT INTO S3 and reform result of SHOW GRANT

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


lance6716:
@GMHDBJD @csuzhangxc PTAL

pkg/check/privilege.go
```diff
@@ -121,10 +121,18 @@ func verifyPrivileges(result *Result, grants []string, expectedGrants []string)
 		firstGrant = firstGrant + " 'secret'"
 	}
 
+	// Aurora has some privilege failing parsing
+	firstGrant = strings.Replace(firstGrant, "LOAD FROM S3", "", 1)
+	firstGrant = strings.ReplaceAll(firstGrant, ", ,", ",")
```
csuzhangxc, line 126: Can we remove this line and only let L128 handle the same logic?

csuzhangxc (commented):
rest LGTM

pkg/check/privilege.go
lance6716: one replacing of this kind is not enough, because
`..., LOAD FROM S3, SELECT INTO S3,...` -> `..., , ,...` -> `..., ,...` ⬅️three adjacent comma will only trigger once.

csuzhangxc (commented):
LGTM

GMHDBJD (approved):
LGTM

lance6716:
@csuzhangxc please approve this PR since we could do that 😂

csuzhangxc (approved):

PR #390, opened 2020-09-11
diff: fix block

WangXiangUSTC:

when killing the process of sync-diff-inspector, diff may firstly exit the `WriteSqls` goroutine, and then will be blocked if have fix SQLs to send to `sqlCh`.

1. use select, add `case <-ctx.Done():`
2. remove the useless `wg.Add(1)` when have fix SQLs send to `sqlCh` by the way, because we can make sure the SQLs which send to `sqlCh` will be handled  before exit `WriteSqls` goroutine.  

### Check List 

Tests 

 - Integration test


WangXiangUSTC:
@csuzhangxc PTAL

csuzhangxc (approved):
LGTM

lance6716:
LGTM

Issue #399, opened 2020-11-05
diff return error `Specified key was too long`


WangXiangUSTC:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.

use diff to check data

2. What did you expect to see?

no error

3. What did you see instead?

get error log
```
[...]

PR #400, opened 2020-11-05
diff: add config MaxIndexLength for tidb

WangXiangUSTC:

fix issue https://github.com/pingcap/tidb-tools/issues/399

set MaxIndexLength to the max value.

### Check List 

Tests 

 - Manual test

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -33,7 +33,6 @@ import (
 	"github.com/pingcap/parser/model"
 	"github.com/pingcap/tidb-tools/pkg/dbutil"
 	"github.com/pingcap/tidb-tools/pkg/utils"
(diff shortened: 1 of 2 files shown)


csuzhangxc:
TiDB also depends on this TiDB-Tools, does this change will affect the global variable used in TiDB itself? 😕 

But if we do not call these codes, it may have no side effects?

WangXiangUSTC:
> TiDB also depends on this TiDB-Tools, does this change will affect the global variable used in TiDB itself? 😕

As long as the TiDB does not execute `setTiDBCfg` will not affect the TiDB's configuration

csuzhangxc (commented):
LGTM

GMHDBJD (approved):
LGTM

csuzhangxc (approved):

PR #402, opened 2020-11-21
add config to ignore tidb stats

3pointer:
when we cannot rely on tidb stats, we need to fail back with randomSpliter to split chunks.

add a config to ignore using tidb stats to split chunks.

### Check List 

Tests 


WangXiangUSTC (commented):
LGTM

lance6716 (approved):

PR #408, opened 2021-01-06
table-filter: turn the package into its own module

kennytm:

This is an experimental change to convert `pkg/table-filter` from a package of tidb-tools to its own submodule. This is intended to reduce the number of unnecessary dependencies (such as "tidb" itself) by other users of table-filter. This also allows us to version this component with semver, independent from TiDB's own versioning scheme.

Moved pkg/table-filter into its own module (to avoid prefix collision). We borrowed the name `components/` from TiKV. The original package is turned into a shim which calls into the component

### Check List 

Tests 

    * Tried to build Dumpling by replacing pkg/table-filter by components/table-filter. Builds fine.

Code changes

Side effects

Related changes


kennytm:
Dumpling's remaining dependency on tidb-tools after table-filter:

* pkg/filter
    * Using `filter.InformationSchemaName` and `filter.PerformanceSchemaName` in tests.
    * Going to replace them with hard-coded constants.
* pkg/utils
    * Using `utils.ToTLSConfig` in config.go
    * Going to extract that into a `components/tls` like this later.
* pkg/dbutil
    * Using `dbutil.IsRetryableError`
    * This is harder to minimize since `IsRetryableError` depends on TiDB's `github.com/pingcap/tidb/errno` package. Ideally the `errno` package could be modularized as well.

components/table-filter/go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/pingcap/tidb-tools/components/table-filter
+
+require (
+	github.com/pingcap/check v0.0.0-20200212061837-5e12011dc712
+	github.com/pingcap/errors v0.11.5-0.20200917111840-a15ef68f753d
+)
+
+go 1.14
```
lance6716, line 8: we use 1.13 other place, should we keep same go version?

lance6716 (commented):
rest LGTM

components/table-filter/go.mod
kennytm: maybe.

kennytm:
It turns out that `tiup` already occupied the name ["components"](https://github.com/pingcap/tiup/tree/master/components) for something else, so we may need to pick another name. How about `modules/`?

PR #415, opened 2021-03-04
schemacmp: set flen and decimal to singleton to avoid dm data mismatch

lichunzhu:
set flen & dec to Singleton.

When dm add a column with different flen, dm can't transfer the sql to modify the column to a larger flen. Change this back to Int after DM support this.

### Check List 

Tests 

 - Unit test


lance6716:
we might need a DM PR using this depedency to run original tests

lichunzhu:
replaced by #416

PR #425, opened 2021-03-12
add online checker

wentaojin:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/sync_diff_inspector/config.go
+++ b/sync_diff_inspector/config.go
@@ -169,6 +169,12 @@ type Config struct {
 	// size of the split chunk
 	ChunkSize int `toml:"chunk-size" json:"chunk-size"`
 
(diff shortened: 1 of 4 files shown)


sync_diff_inspector/config.go
```diff
@@ -299,6 +305,15 @@ func (c *Config) checkConfig() bool {
 		return false
 	}
 
+	if c.FailedRetryTimes <= 0 {
```
lance6716, line 308: we might set default value for them, or allow `0` failed retry

wentaojin: allowing 0  may be better

PR #426, opened 2021-03-17
schemacmp: keep column flag consistent with MySQL

sleepymole:

From TiDB [ddl_api](https://github.com/pingcap/tidb/blob/080ed4ee7b5b492fcc17bdc881100f3478aa7118/ddl/ddl_api.go#L292), if a index has multiple columns, only the first column can be applied the MultipleKeyFlag. However, schemacmp.Join applies flag to all the index columns. It causes some strange behavior. For example,
```golang
a := toTableInfo("create table t(a int primary key, b int, c int, d int, e int, f int, unique key uni_bc(b,c), unique key uni_d(d), index idx_ef(e,f))")
b := toTableInfo("create table t(a int primary key, b int, c int, d int, e int, f int, unique key uni_bc(b,c), unique key uni_d(d), index idx_ef(e,f))")
aa := schemacmp.Encode(a)
bb := schemacmp.Encode(b)
joined, _ := aa.Join(bb)
cmp, err := aa.Compare(joined)
assert(cmp == 0 && err == nil) // failed
```
This pr tries to  keep column flag consistent with TiDB & MySQL after joined, so that the joined table of two same table is equal to the original table.

Keep column flag consistent with MySQL after joined.

### Check List 

Tests 

 - Unit test
 
Code changes

Side effects

Related changes

### DIFF
--- a/pkg/schemacmp/table.go
+++ b/pkg/schemacmp/table.go
@@ -386,21 +386,21 @@ func (t Table) Join(other Table) (Table, error) {
 	columnKeyFlags := make(map[string]uint)
 	table := res.(Tuple)
 	for _, index := range table[tableInfoTupleIndexIndices].(latticeMap).LatticeMap.(indexMap) {
-		var flag uint
+		cols := index[indexInfoTupleIndexColumns].Unwrap().(indexColumnSlice)
+		if len(cols) == 0 {
+			continue
+		}
 		switch {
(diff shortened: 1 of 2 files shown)


lance6716 (commented):
/lgtm

kennytm:
LGTM

lichunzhu (approved):

PR #429, opened 2021-04-01
schemacmp: skip empty charset and collate in table.String()

sleepymole:

When table's charset or collate is empty, table.String() still writes charset or collate keyword, such as `CREATE TABLE tbl (id INT) CHARSET  COLLATE`, which is not a valid sql syntax.

### What is changed and how it works?i

Skip charset or collate keyword when it is empty.

### Check List 

Tests 

 - Unit test
 
Code changes

Side effects

Related changes

### DIFF
--- a/pkg/schemacmp/table_test.go
+++ b/pkg/schemacmp/table_test.go
@@ -530,6 +530,8 @@ func (s *tableSchema) TestTableString(c *C) {
 			sql := strings.ToLower(schemacmp.Encode(ti).String())
 			c.Assert(strings.Contains(sql, "charset"), Equals, charset != "")
 			c.Assert(strings.Contains(sql, "collate"), Equals, collate != "")
+			_, err := s.toTableInfo(sql)
+			c.Assert(err, IsNil)
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


pkg/schemacmp/table_test.go
```diff
@@ -513,3 +516,20 @@ func (s *tableSchema) TestJoinSchemas(c *C) {
...
+	ti, err := s.toTableInfo("CREATE TABLE tb (a INT, b INT)")
+	c.Assert(err, IsNil)
+
+	charsets := []string{"", mysql.DefaultCharset}
+	collates := []string{"", mysql.DefaultCollationName}
+	for _, charset := range charsets {
+		for _, collate := range collates {
+			ti.Charset = charset
+			ti.Collate = collate
+			sql := strings.ToLower(schemacmp.Encode(ti).String())
```
lance6716, line 530: we might test this SQL could be successfully parsed

lance6716 (commented):
rest lgtm

pkg/schemacmp/table_test.go
sleepymole: done

PR #430, opened 2021-04-15
check: pretty print the checker errors

lance6716:

https://asktug.com/t/topic/69143

example

```
{
	"id": 0,
	"name": "table structure compatibility check",
	"desc": "check compatibility of table structure",
	"state": "fail",
	"errors": [
		{
			"severity": "warn",
			"short_error": "table `test-db`.`test-table-1` Foreign Key fk is parsed but ignored by TiDB.",
			"instruction": "please ref document: https://docs.pingcap.com/tidb/stable/mysql-compatibility#unsupported-features"
		},
		{
			"severity": "fail",
			"short_error": "table `test-db`.`test-table-1` primary/unique key does not exist",
			"instruction": "please set primary/unique key for the table"
		}
	],
	"extra": "address of db instance - :0"
}
```

```
{
		"id": 0,
		"name": "sharding table test-name consistency checking",
		"desc": "check consistency of sharding table structures",
		"state": "fail",
		"errors": [
			{
				"severity": "fail",
				"short_error": "column length mismatch (1 vs 2)",
				"left": "instance test-source table `test-db`.`test-table-1` columns [c]",
				"right": "instance test-source table `test-db`.`test-table-2` columns [c d]"
			}
		],
		"instruction": "please set same table structure for sharding tables",
		"extra": "error on sharding test-name"
	}
```

### Check List 

Tests 

 - Unit test

### DIFF
--- a/pkg/check/check.go
+++ b/pkg/check/check.go
@@ -40,8 +40,8 @@ const (
 type Error struct {
 	Severity    State  `json:"severity"`
 	ShortErr    string `json:"short_error"`
-	Left        string `json:"left,omitempty"`
-	Right       string `json:"right,omitempty"`
+	Self        string `json:"self,omitempty"`
+	Other       string `json:"other,omitempty"`
 	Instruction string `json:"instruction,omitempty"`
 }
 
--- a/pkg/check/check.go
+++ b/pkg/check/check.go
@@ -37,15 +37,27 @@ const (
 	StateWarning State = "warn"
 )
 
+type Error struct {
+	Severity    State  `json:"severity"`
+	ShortErr    string `json:"short_error"`
+	Left        string `json:"left,omitempty"`
+	Right       string `json:"right,omitempty"`
+	Instruction string `json:"instruction,omitempty"`
+}
+
+func NewError(description string) *Error {
+	return &Error{Severity: StateFailure, ShortErr: description}
+}
+
 // Result is result of check
 type Result struct {
-	ID          uint64 `json:"id"`
-	Name        string `json:"name"`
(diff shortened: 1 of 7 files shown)


pkg/check/check.go
```diff
@@ -37,15 +37,27 @@ const (
 	StateWarning State = "warn"
 )
 
+type Error struct {
+	Severity    State  `json:"severity"`
+	ShortErr    string `json:"short_error"`
+	Left        string `json:"left,omitempty"`
```
GMHDBJD: use a meaningful name instead of left and right in error message?

GMHDBJD (commented):
rest LGTM

pkg/check/check.go
lance6716: that's corresponding to `column length mismatch (1 vs 2)`. How about self and other?

column length mismatch (self: 1 vs other: 2)

GMHDBJD: make sense

pkg/check/check.go
```diff
@@ -37,15 +37,27 @@ const (
...
 
+type Error struct {
+	Severity    State  `json:"severity"`
+	ShortErr    string `json:"short_error"`
+	Self        string `json:"self,omitempty"`
+	Other       string `json:"other,omitempty"`
+	Instruction string `json:"instruction,omitempty"`
+}
+
+func NewError(description string) *Error {
```
lichunzhu: What about using `NewError(description string, ...args)`? These `fmt.Sprintf` look strange 🤔

lichunzhu (commented):
Rest LGTM

PR #431, opened 2021-04-16
schemacmp: support partition

sleepymole:

Support comparing and joining table with partition in schemacmp.

Encode partition info into `schemacmp.Table` and compare it.

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes

### DIFF
--- a/pkg/schemacmp/lattice.go
+++ b/pkg/schemacmp/lattice.go
@@ -469,6 +469,42 @@ func (a Uint) Join(other Lattice) (Lattice, error) {
 	}
 }
 
+// Uint64 is a uint implementing Lattice.
+type Uint64 uint
+
+// Unwrap implements Lattice.
+func (a Uint64) Unwrap() interface{} {
+	return uint64(a)
+}
+
+// Compare implements Lattice.
+func (a Uint64) Compare(other Lattice) (int, error) {
+	b, ok := other.(Uint64)
+	switch {
+	case !ok:
+		return 0, typeMismatchError(a, other)
+	case a == b:
+		return 0, nil
+	case a > b:
+		return 1, nil
+	default:
+		return -1, nil
+	}
+}
+
+// Join implements Lattice.
+func (a Uint64) Join(other Lattice) (Lattice, error) {
+	b, ok := other.(Uint64)
+	switch {
+	case !ok:
+		return nil, typeMismatchError(a, other)
+	case a >= b:
+		return a, nil
+	default:
+		return b, nil
+	}
+}
+
 // Tuple of Lattice instances. Given two Tuples `a` and `b`, we define `a <= b`
 // iff `a[i] <= b[i]` for all `i`.
 type Tuple []Lattice
@@ -609,7 +645,7 @@ func (a maybe) Join(other Lattice) (Lattice, error) {
 	}
 }
 
-// StringList is a list of string where `a <= b` iff `a == b[:len(a)]`.
+// StringList is a list of string where `a <= b` if `a == b[:len(a)]`.
 type StringList []string
 
 // Unwrap implements Lattice.
--- a/pkg/schemacmp/lattice.go
+++ b/pkg/schemacmp/lattice.go
@@ -470,7 +470,7 @@ func (a Uint) Join(other Lattice) (Lattice, error) {
 }
 
 // Uint64 is a uint implementing Lattice.
-type Uint64 uint
+type Uint64 uint64
 
 // Unwrap implements Lattice.
 func (a Uint64) Unwrap() interface{} {
--- a/pkg/schemacmp/table.go
+++ b/pkg/schemacmp/table.go
@@ -319,7 +319,6 @@ const (
 	partitionTupleIndexType = iota
 	partitionTupleIndexExpr
 	partitionTupleIndexColumns
-	partitionTupleIndexEnable
 	partitionTupleIndexDefinitions
 	partitionTupleIndexNum
 )
--- a/pkg/schemacmp/table.go
+++ b/pkg/schemacmp/table.go
@@ -373,7 +373,6 @@ func restorePartitionInfoFromUnwrapped(ctx *format.RestoreCtx, info []interface{
 				ctx.WritePlain(", ")
 			}
 			ctx.WriteKeyWord("PARTITION ")
-			// ctx.WritePlain(pair.key)
 			ctx.WriteName(pair.key)
 
 			d := pair.value.(partitionDefinition)
--- a/pkg/schemacmp/table.go
+++ b/pkg/schemacmp/table.go
@@ -235,6 +235,169 @@ func (indexMap) JoinWithNil(value Lattice) (Lattice, error) {
 	return nil, nil
 }
 
+type partitionDefinition struct {
+	lessThan []string
+	inValues [][]string
+}
+
+func (a partitionDefinition) Equals(other Equality) bool {
+	b, ok := other.(partitionDefinition)
+	if !ok {
+		return false
+	}
+	if len(a.lessThan) != len(b.lessThan) {
+		return false
+	}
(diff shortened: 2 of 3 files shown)


pkg/schemacmp/lattice.go
```diff
@@ -469,6 +469,42 @@ func (a Uint) Join(other Lattice) (Lattice, error) {
 	}
 }
 
+// Uint64 is a uint implementing Lattice.
+type Uint64 uint
```
lance6716: uint64 🤔

pkg/schemacmp/table.go
```diff
@@ -235,6 +235,168 @@ func (indexMap) JoinWithNil(value Lattice) (Lattice, error) {
...
+	num := Singleton(pi.Num)
+	// Only range and list partition num can be changed.
+	if pi.Type == model.PartitionTypeRange || pi.Type == model.PartitionTypeList {
+		num = Uint64(pi.Num)
+	}
+	return Tuple{
+		Singleton(pi.Type),
+		Singleton(pi.Expr),
+		columns,
+		Singleton(pi.Enable),
```
lance6716: by reading the comment of `pi.Enable`, I think this is a TiDB specified attribute so we don't need it.

lance6716 (commented):
what's the behaviour for https://dev.mysql.com/doc/refman/5.7/en/partitioning-subpartitions.html

sleepymole:
> what's the behaviour for https://dev.mysql.com/doc/refman/5.7/en/partitioning-subpartitions.html

I didn't find subpartitioning in model.TableInfo. I will check it again later.

pkg/schemacmp/lattice.go
sleepymole: thanks, this is a typo

pkg/schemacmp/table.go
```diff
@@ -235,6 +235,168 @@ func (indexMap) JoinWithNil(value Lattice) (Lattice, error) {
...
+		ctx.WritePlain(" (")
+		for i, pair := range sortedMap(info[partitionTupleIndexDefinitions].(map[string]interface{})) {
+			if i > 0 {
+				ctx.WritePlain(", ")
+			}
+			ctx.WriteKeyWord("PARTITION ")
+			ctx.WriteName(pair.key)
+
+			d := pair.value.(partitionDefinition)
+			if len(d.lessThan) > 0 {
```
lance6716, line 382: in create table AST, value less than could be an expression. So here did it evaluate and only save the result?

pkg/schemacmp/table_test.go
```diff
@@ -518,20 +608,54 @@ func (s *tableSchema) TestJoinSchemas(c *C) {
...
 		}
 	}
+
+	// Table with partition should be parsed after encoded.
+
+	// Range partition.
+	ti, err = s.toTableInfo("CREATE TABLE tb (id INT NOT NULL) PARTITION BY RANGE (id) (PARTITION p0 VALUES LESS THAN (10), PARTITION p1 VALUES LESS THAN (20))")
+	c.Assert(err, IsNil)
+	sql := strings.ToLower(Encode(ti).String())
+	c.Assert(strings.Contains(sql, "partition by range"), Equals, true)
```
lance6716, line 635: I think above line we use ToLower to handle `charset` and `CHARSET`, here no need to use it

pkg/schemacmp/table.go
sleepymole: yes, table info only saved the string results.

pkg/schemacmp/table_test.go
sleepymole: here is to check whether the sql  statement contains the partition definition.

lance6716:
we should also forbid "using ADD to rollback a partitially dropped partition" 😭 
and REORGANIZE is different from ADD, we might permit reorg? 😵 

 https://dev.mysql.com/doc/refman/5.7/en/partitioning-management-range-list.html

pkg/schemacmp/table_test.go
lance6716: We could remove ToLower and check PARTITION BY RANGE

sleepymole:
> we should also forbid "using ADD to rollback a partially dropped partition" 😭

If we just remove a partition, I think the upstream should make sure all data in the partition should be deleted. And there is a `TRUNCATE PARTITION`  operation which is similar with `DROP COLUMN`.  But it seems `TRUNCATE` doesn't change the table structure. So we can't do much in schemacmp.

> and REORGANIZE is different from ADD, we might permit reorg? 😵

TiDB doesn't support reorg currently. And there are more operations which are not supported by TiDB. If the downstream is always TiDB, I think we can forbid these operations.
> With regard to partition management, any operation that requires moving data in the bottom implementation is not supported currently, including but not limited to: adjust the number of partitions in a Hash partitioned table, modify the Range of a Range partitioned table, merge partitions and exchange partitions.

pkg/schemacmp/table_test.go
sleepymole: `charset` check is no need to use `ToLower` too 😂

lance6716: maybe the original auther wants to check `charset` and `CHARSET`

sleepymole: I'm the original author 😂

sleepymole: I originally wanted to increase the stability, because the SQL keywords are case-insensitive.

sleepymole:
> > what's the behaviour for https://dev.mysql.com/doc/refman/5.7/en/partitioning-subpartitions.html
> 
> I didn't find subpartitioning in model.TableInfo. I will check it again later.

@lance6716  Subpartitioning are only supported in parser. TiDB doesn't support subpartitioning  currently and there is no subpartitioning  info in `model.TableInfo`.

pkg/schemacmp/table.go
```diff
@@ -235,6 +235,168 @@ func (indexMap) JoinWithNil(value Lattice) (Lattice, error) {
...
+		av := a.inValues[i]
+		bv := b.inValues[i]
+		if len(av) != len(bv) {
+			return false
+		}
+		for j := 0; j < len(av); j++ {
+			if av[j] != bv[j] {
+				return false
+			}
+		}
```
lichunzhu, line 269: We can write a sub-function for this part and L248-255

lance6716:
Now we know that DROP PARTITION will delete the data in that PARTITION.

for two sharding tables t1, t2 with same partition [MIN_VALUE, 100), [100, 200). consider this case:

1. t1 DROP PARTITION [MIN_VALUE, 100), now the only partition is [MIN_VALUE, 200). the joined didn't change, so no DDL executed
2. t1 INSERT 50
3. t2 DROP PARTITION [MIN_VALUE, 100). the joined changed, downstream DROP PARTITION [MIN_VALUE, 100)

but downstream will lost INSERT 50 in step 2 🤔 maybe optimistic sharding can't support PARTITION

sleepymole:
> Now we know that DROP PARTITION will delete the data in that PARTITION.
> 
> for two sharding tables t1, t2 with same partition [MIN_VALUE, 100), [100, 200). consider this case:
> 
> 1. t1 DROP PARTITION [MIN_VALUE, 100), now the only partition is [MIN_VALUE, 200). the joined didn't change, so no DDL executed
> 2. t1 INSERT 50
> 3. t2 DROP PARTITION [MIN_VALUE, 100). the joined changed, downstream DROP PARTITION [MIN_VALUE, 100)
> 
> but downstream will lost INSERT 50 in step 2 🤔 maybe optimistic sharding can't support PARTITION

too sad ...

sleepymole:
As the discussion above, it seems there is no way that optimistic sharding will support PARTITION. @lance6716  How about close this PR.

lance6716:
(closing, thanks)

PR #432, opened 2021-04-20
dbutil: fix can't handle masked password in SHOW GRANTS

lance6716:
https://asktug.com/t/topic/69695/3?u=lance6716

move strings.Replace `<secret>` to an earlier place

### Check List 

Tests 

 - Unit test


GMHDBJD (approved):
/lgtm

PR #440, opened 2021-05-21
*: bump go to 1.16

Ehco1996:

try bump go to 1.16 so we can build this tools on mac m1 chip #437 

update go mod to 1.1.6

### Check List 

Tests 

 - Unit test

Code changes

- None

Side effects

- None

Related changes

- None


kennytm:
The `go.mod1` files are used to break the cyclic dependency between pingcap/tidb and this module. I don't think upgrading to Go 1.16 is going to allow us renaming it back to `go.mod`.

Ehco1996:
> The `go.mod1` files are used to break the cyclic dependency between pingcap/tidb and this module. I don't think upgrading to Go 1.16 is going to allow us renaming it back to `go.mod`.

know it

PR #441, opened 2021-06-03
filter: add another Apply function for filter to return the original value

glorv:
- Add a new `Filter.Apply` method to avoid change the original schema/table name to lower case in case-insensitive mode
- deprecate the `Filter.ApplyOn` method, caller should not depend on the side effect

### Check List 

Tests 

 - Unit test

Code changes

Side effects

Related changes


lichunzhu (approved):
LGTM

lance6716 (approved):

PR #442, opened 2021-06-11
sync-diff: fix the problem that sync diff data is null

lichunzhu:
When some return data from mysql is null sync-diff will fail.

sync-diff: fix the problem that sync diff data is null

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

### DIFF
--- a/pkg/dbutil/common.go
+++ b/pkg/dbutil/common.go
@@ -235,12 +235,14 @@ func GetRandomValues(ctx context.Context, db *sql.DB, schemaName, table, column 
 
 	randomValue := make([]string, 0, num)
 	for rows.Next() {
-		var value string
+		var value sql.NullString
 		err = rows.Scan(&value)
 		if err != nil {
 			return nil, errors.Trace(err)
 		}
-		randomValue = append(randomValue, value)
+		if value.Valid {
+			randomValue = append(randomValue, value.String)
+		}
 	}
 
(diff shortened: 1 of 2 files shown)


lance6716:
I'll add integration tests in anothe PR.

manual test 1:

```sql
CREATE TABLE `t` (
  `c` int(11) DEFAULT NULL,
  UNIQUE KEY `c` (`c`)
)
```

```sql
MySQL [test]> select * from t;
+------+
| c    |
+------+
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
| NULL |
|    1 |
|    2 |
|    3 |
+------+
29 rows in set (0.003 sec)
```

set `chunk-size = 5` in config

lance6716:
manual test 2:

```sql
MySQL [test]> select * from t;
+------+
| c    |
+------+
|    1 |
|    2 |
+------+
2 rows in set (0.003 sec)

MySQL [test]> select * from t2;
+------+
| c    |
+------+
| NULL |
|    2 |
+------+
2 rows in set (0.002 sec)
```

lance6716 (approved):

PR #447, opened 2021-06-21
sync_diff_inspector: make exclude table support regex

lichunzhu:
When we try to exclude some tables, we need to write them one by one which is inconvenient.

make `check-tables.exclude_table` support regex

### Check List 

Tests 

 - Unit test
 - Integration test
```toml
[[check-tables]]
schema = "chauncy"
tables = ["~.*"]
exclude-tables = ["~t[2-3]"]
```
Run with this config and t2, t3 are not equal. diff can pass.

Related changes

### DIFF
--- a/sync_diff_inspector/diff.go
+++ b/sync_diff_inspector/diff.go
@@ -477,11 +477,11 @@ func (df *Diff) adjustDBCfgByDMSubTasks(cfg *Config) error {
 }
 
 // GetAllTables get all tables in all databases.
-func (df *Diff) GetAllTables(cfg *Config) (map[string]map[string]map[string]interface{}, error) {
+func (df *Diff) GetAllTables(cfg *Config) (map[string]map[string]map[string]struct{}, error) {
 	// instanceID => schema => table
-	allTablesMap := make(map[string]map[string]map[string]interface{})
+	allTablesMap := make(map[string]map[string]map[string]struct{})
 
-	allTablesMap[df.targetDB.InstanceID] = make(map[string]map[string]interface{})
+	allTablesMap[df.targetDB.InstanceID] = make(map[string]map[string]struct{})
 	targetSchemas, err := dbutil.GetSchemas(df.ctx, df.targetDB.Conn)
 	if err != nil {
 		return nil, errors.Annotatef(err, "get schemas from %s", df.targetDB.InstanceID)
@@ -495,7 +495,7 @@ func (df *Diff) GetAllTables(cfg *Config) (map[string]map[string]map[string]inte
 	}
 
 	for _, source := range df.sourceDBs {
-		allTablesMap[source.InstanceID] = make(map[string]map[string]interface{})
+		allTablesMap[source.InstanceID] = make(map[string]map[string]struct{})
 		sourceSchemas, err := dbutil.GetSchemas(df.ctx, source.Conn)
 		if err != nil {
 			return nil, errors.Annotatef(err, "get schemas from %s", source.InstanceID)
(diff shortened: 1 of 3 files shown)


sync_diff_inspector/diff.go
```diff
@@ -326,12 +326,11 @@ func (df *Diff) AdjustTableConfig(cfg *Config) (err error) {
...
-				if df.InExcludeTables(schemaTables.ExcludeTables, t) {
-					continue
-				} else {
-					tables = append(tables, t)
-				}
+			for _, excludeTable := range schemaTables.ExcludeTables {
+				matchedTables = df.RemoveMatchTable(excludeTable, matchedTables)
+			}
+			for table := range matchedTables {
+				tables = append(tables, table)
```
glorv, line 336:
```suggestion
			tables = append(tables, matchedTables...)
```

sync_diff_inspector/diff.go
```diff
@@ -537,6 +536,25 @@ func (df *Diff) GetMatchTable(db DBConfig, schema, table string, allTables map[s
 	return tableNames, nil
 }
 
+// RemoveMatchTable removes all the matched table.
+func (df *Diff) RemoveMatchTable(table string, allTables map[string]interface{}) map[string]interface{} {
```
glorv: I think we don't need to depend on an `map[string]interface{}` only to do this remove operation, just use slice can do the same thing.

sync_diff_inspector/diff.go
glorv: If we don't change matchTables from  `slice` to `map`

sync_diff_inspector/diff.go
```diff
@@ -537,6 +536,25 @@ func (df *Diff) GetMatchTable(db DBConfig, schema, table string, allTables map[s
...
+	tableNames := allTables
+
+	if table[0] == '~' {
+		tableRegex := regexp.MustCompile(fmt.Sprintf("(?i)%s", table[1:]))
+		for tableName := range tableNames {
+			if !tableRegex.MatchString(tableName) {
```
lance6716: if match {delete}

sync_diff_inspector/diff.go
lance6716: I prefer `map[string]struct{}`

lichunzhu: `map[string]struct{}` is easier to delete.

lance6716 (approved):

sync_diff_inspector/diff.go
glorv: LGTM，since the original structure is already map not slice

sync_diff_inspector/diff.go
```diff
@@ -537,6 +535,24 @@ func (df *Diff) GetMatchTable(db DBConfig, schema, table string, allTables map[s
 	return tableNames, nil
 }
 
+// RemoveMatchTable removes all the matched table.
+func (df *Diff) RemoveMatchTable(table string, allTables map[string]struct{}) map[string]struct{} {
+	tableNames := allTables
+
+	if table[0] == '~' {
+		tableRegex := regexp.MustCompile(fmt.Sprintf("(?i)%s", table[1:]))
```
glorv: How can we ensure the regexp patten is always a valid expression

glorv: BTW, do we need to ensure the expression should match the whole table name by alway add "^$" to the expression？

lichunzhu: > How can we ensure the regexp patten is always a valid expression

sync-diff will panic directly in the past. Now I have updated this part to report an error

lichunzhu: > BTW, do we need to ensure the expression should match the whole table name by alway add "^$" to the expression？

I think we should. But dm doesn't have this rule either.
Maybe we could have a further discussion about this and do this in another PR? Current configuration doesn't violate the situation before: https://github.com/pingcap/tidb-tools/blob/915b22e4d42c1493edf90f1ccc79227e55e217ba/sync_diff_inspector/diff.go#L522

lichunzhu:
Sync-diff 2.0 replace filter with `table_filter`, so this PR can be closed now.

PR #448, opened 2021-06-21
sync_diff_test: change mydumper to dumpling

lichunzhu:
change mydumper in sync_diff test to dumpling 

change mydumper in sync_diff test to dumpling 

### Check List 

Tests 

 - Unit test
 - Integration test


lance6716:
need change CI to download dumpling binary

PR #450, opened 2021-07-06
sync_diff_inspector: optimize heap load for multi-sources

Leavrth:

It is not necessary to traverse all sources whenever it obtains one row.

Just keep the heap saving one row for each source, otherwise the source has read to the end.
When the heap pops one row, it knows which source this row is from. So whenever the heap pops one row, push new one row from the source which the popped row is from.

### Check List 

Related changes

### DIFF
--- a/pkg/diff/diff.go
+++ b/pkg/diff/diff.go
@@ -557,7 +557,6 @@ func (t *TableDiff) compareRows(ctx context.Context, chunk *ChunkRange) (bool, e
 	beginTime := time.Now()
 
 	sourceRows := make(map[int]*sql.Rows)
-	sourceHaveData := make(map[int]bool)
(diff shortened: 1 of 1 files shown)


lance6716:
the original code seems a lazy reading from each sources, could you explain what's the benifits of your changing?

Leavrth:
> the original code seems a lazy reading from each sources, could you explain what's the benifits of your changing?

This function only returns one row each time, so I think it's better to record only the source which the last popped row is from. If we need a lazy reading, I think a variable `lastRowSource` is enough instead of array `sourceHaveData`.

3pointer (approved):
LGTM, nice catch!

Leavrth:
Due to sync_diff_inspector has been updated to version 2.0. This PR is outdated.

PR #456, opened 2021-07-22
abstract sql.DB to interface 

glorv:
…

add a interface to allow call some util function with both *sql.DB and *sql.Conn.

### Check List 

Tests 

 - Unit test
 - Integration test

Code changes

   - add context.Context parameter to some functions 
 - Has persistent data change

Side effects

Related changes


lance6716 (approved):

lichunzhu (approved):
LGTM

PR #469, opened 2021-08-04
diff2: add test

Leavrth:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


sync_diff_inspector/source/basic.go
```diff
@@ -32,6 +32,7 @@ import (
 type BasicSource struct {
 	tableDiffs []*common.TableDiff
 	dbConn     *sql.DB
+	ctx        context.Context
```
3pointer: According to https://pkg.go.dev/context,  store `ctx` in struct seems not a good idea?

PR #485, opened 2021-08-20
Diff 2.0: impl summary

ZipFast:
### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


sync_diff_inspector/source/basic.go
```diff
@@ -99,9 +98,19 @@ func getOriginTable(sourceTableMap map[string]*common.TableSource, table *common
 	return originSchema, originTable
 }
 
+func (s *BasicSource) GetSourceTableMap() map[string]*common.TableSource {
```
3pointer: why add this function?

sync_diff_inspector/utils/utils.go
```diff
@@ -58,6 +59,45 @@ func NewWorkerPool(limit uint, name string) *WorkerPool {
 	}
 }
 
+// Returns either an ascii code, or (if input is an arrow) a Javascript key code.
+func GetChar() (ascii int, keyCode int, err error) {
```
3pointer, line 64: maybe we should ask PM to add this interactive action🤔

sync_diff_inspector/diff.go
```diff
@@ -152,23 +205,77 @@ func (df *Diff) Equal(ctx context.Context) error {
...
+		log.Info("generate chunk", zap.Int("chunk id", c.ChunkRange.ID), zap.String("table", dbutil.TableName(tableDiff.Schema, tableDiff.Table)))
 		pool.Apply(func() {
-			res, err := df.consume(ctx, c)
-			if err != nil {
-				// TODO: catch error
-			}
-			// TODO: handle res
-			if res {
-
+			tableDiff := df.workSource.GetTables()[c.TableIndex]
```
3pointer: this function needs a lock.
I suggest save the `schema` `table` into Chunk

sync_diff_inspector/source/basic.go
3pointer: We can implement the `GetTableStruct` in each Source. `SourceTableMap` is too implied

ZipFast: Fixed

sync_diff_inspector/diff.go
ZipFast: I cannot see there is any concurrent writes here

sync_diff_inspector/diff.go
```diff
@@ -152,23 +205,64 @@ func (df *Diff) Equal(ctx context.Context) error {
...
+				df.report.SetTableMeetError(schema, table, err)
 			}
 		})
 	}
 
 	return nil
 }
 
+func (df *Diff) compareStruct(ctx context.Context, c *splitter.RangeInfo) (structEq bool, err error) {
+	switch upstream := df.upstream.(type) {
```
3pointer: This code can be refine simplify
```
sourceTableInfos, err := upstream.GetSourceStructInfo(ctx, c.TableIndex)
if err != nil {
    return false, errors.Trace(err)
}
structEq = true
for _, tableInfo := range sourceTableInfos {
    eq, _ := dbutil.EqualTableInfo(tableInfo, df.downstream.GetTables()[c.TableIndex].Info)
    structEq = structEq && eq
}
```

sync_diff_inspector/diff.go
```diff
@@ -199,7 +293,10 @@ func (df *Diff) generateChunksIterator(ctx context.Context) (source.RangeIterato
 					zap.Int("id", node.GetID()),
 					zap.Reflect("chunk", node),
 					zap.String("state", node.GetState()))
-				df.cp.SetCurrentSavedID(node.GetID() + 1)
+				df.cp.SetCurrentSavedID(node.GetID())
+			}
+			if err = df.loadReport(path + "_report"); err != nil {
```
3pointer, line 289: 🤔 we'd better put the report at the same level of checkpoint directory. to compare with checkpoint. report is an Independent part.

sync_diff_inspector/diff.go
```diff
@@ -152,23 +205,64 @@ func (df *Diff) Equal(ctx context.Context) error {
...
-			if res {
-
+			tableDiff := df.workSource.GetTables()[c.TableIndex]
+			schema, table := tableDiff.Schema, tableDiff.Table
+			structEq, err := df.compareStruct(ctx, c)
+			if err == nil {
+				if !structEq {
+					df.report.SetTableStructCheckResult(schema, table, structEq)
+				}
+				res, rowsAdd, rowsDelete, rowsCnt, err := df.consume(ctx, c)
```
3pointer: config `ignore-data-check` and `ignore-struct-check` can be add here.

sync_diff_inspector/diff.go
3pointer: ok, There is no race.

but In this code we will repeated compare the same table `chunksCount` times. we need reduce the compare times. 🤔

sync_diff_inspector/diff.go
ZipFast: this report should be a temp file, and will be deleted in the end of diff process

3pointer: yes, but it's strange to see a file called `checkpoint.pb_report` that has nothing to do with checkpoint.e

PR #520, opened 2021-09-27
support table transfer column for dm

yufan022:
part of: https://github.com/pingcap/dm/pull/2178

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes


lance6716:
I'm checking if our CLA checker is online, will reopen soon.

pkg/table-router/router.go
```diff
@@ -23,10 +24,17 @@ import (
...
+	TablePattern        string               `json:"table-pattern" toml:"table-pattern" yaml:"table-pattern"`
+	TargetSchema        string               `json:"target-schema" toml:"target-schema" yaml:"target-schema"`
+	TargetTable         string               `json:"target-table" toml:"target-table" yaml:"target-table"`
+	TableTransferColumn *TableTransferColumn `json:"table-transfer-column" toml:"table-transfer-column" yaml:"table-transfer-column"`
+}
+
+// TableTransferColumn is a rule that converts table names to column values
+type TableTransferColumn struct {
+	TargetColumn string `json:"target-column" toml:"target-column" yaml:"target-column"`
+	TableRegular string `json:"table-regular" toml:"table-regular" yaml:"table-regular"`
```
lance6716: this name is confusing, how about rename it to `extract-table`, `table-extractor`

> There are only two hard things in Computer Science: cache invalidation and naming things.

BTW, would you like to add a `schema-extractor` even if you don't use it? (and we'll find a good place to add `source-extractor` in future)

yufan022:
```
type TableRule struct {
	SchemaPattern   string           `json:"schema-pattern" toml:"schema-pattern" yaml:"schema-pattern"`
	TablePattern    string           `json:"table-pattern" toml:"table-pattern" yaml:"table-pattern"`
	TargetSchema    string           `json:"target-schema" toml:"target-schema" yaml:"target-schema"`
	TargetTable     string           `json:"target-table" toml:"target-table" yaml:"target-table"`
	TableExtractor  *TableExtractor  `json:"extract-table" toml:"extract-table" yaml:"extract-table"`
	SchemaExtractor *SchemaExtractor `json:"extract-schema" toml:"extract-schema" yaml:"extract-schema"`
	SourceExtractor *SourceExtractor `json:"extract-source" toml:"extract-source" yaml:"extract-source"`
}

// TableExtractor is a rule that converts table names to column values
type TableExtractor struct {
	TargetColumn string `json:"target-column" toml:"target-column" yaml:"target-column"`
	TableRegular string `json:"table-regular" toml:"table-regular" yaml:"table-regular"`
}
type SchemaExtractor struct {
	TargetColumn  string `json:"target-column" toml:"target-column" yaml:"target-column"`
	SchemaRegular string `json:"schema-regular" toml:"table-regular" yaml:"table-regular"`
}
type SourceExtractor struct {
	TargetColumn  string `json:"target-column" toml:"target-column" yaml:"target-column"`
	SourceRegular string `json:"source-regular" toml:"source-regular" yaml:"source-regular"`
}

```

How does it look now? Maybe schema and source also need to be extracted into columns, is that your means? @lance6716

tisonkun:
testing cla-assistant

pkg/table-router/router.go
Ehco1996: how about calling `TableTransferColumn `  `TargetVirtualShardingKey`

pkg/table-router/router_test.go
```diff
@@ -30,11 +30,11 @@ type testRouterSuite struct{}
...
-		{"Test_1_*", "abc*", "t1", "abc"},
-		{"test_1_*", "test*", "t2", "test"},
-		{"test_1_*", "", "test", ""},
-		{"test_2_*", "abc*", "t1", "abc"},
-		{"test_2_*", "test*", "t2", "test"},
+		{SchemaPattern: "Test_1_*", TablePattern: "abc*", TargetSchema: "t1", TargetTable: "abc"},
+		{SchemaPattern: "test_1_*", TablePattern: "test*", TargetSchema: "t2", TargetTable: "test"},
+		{SchemaPattern: "test_1_*", TablePattern: "", TargetSchema: "test", TargetTable: ""},
+		{SchemaPattern: "test_2_*", TablePattern: "abc*", TargetSchema: "t1", TargetTable: "abc"},
+		{SchemaPattern: "test_2_*", TablePattern: "test*", TargetSchema: "t2", TargetTable: "test"},
```
Ehco1996, line 37: please add some test for new added fields

pkg/table-router/router.go
lance6716: @yufan022 

> Maybe schema and source also need to be extracted into columns, is that your means?

That's a good design, my original idea is to handle this case:

user may need to attach extra column by its source id, or dabatase name, or table name. For example, when merging `us-west.user` table and `us-east.user` table, the database name may need to be attached. 

My original idea is there's only one extra column, and the format maybe configurated such as `{source-extractor}-{schema-extractor}-{table-extractor}`. Your idea is add (up to) three extra columns, I think that's better because we don't need to write such a complicated formatter string. 👍

>  How does it look now?

the `regular` in `table-regular` seems unfamiliar to me, maybe other users won't recognize it as regular expression, can you give this usage in other software configurations? BTW I prefer `table-regexp`

And above all, if you just want to make a quick contribution and don't want to pay more time at over-design for now, we can move forward!

@Ehco1996 

> how about calling TableTransferColumn TargetVirtualShardingKey

that's too long, I don't like fields more than two words 😂

Ehco1996: >that's too long, I don't like fields more than two word

ok, we can find anothor name for this field

yufan022: > the regular in table-regular seems unfamiliar to me, maybe other users won't recognize it as regular expression, can you give this usage in other software configurations? BTW I prefer table-regexp

`table-regexp` looks better.

> And above all, if you just want to make a quick contribution and don't want to pay more time at over-design for now, we can move forward!

I'm glad to see this feature more standardized, So you can put forward more ideas, and I will try to improve it.

sunetic: I actually prefer to generate one column rather than three. The newly added column acts as an identifier rather than part of the business data, and more columns may be too intrusive for the user to bother with

Regarding the naming, I am nicknamed "The Black Hole of Naming", my opinion is for reference.

First of all, it must be very clearly distinguished from the existing parameters, i.e. `target-table`, `target-schema`. Because of the secondary frequency of use, must not be similar to the common parameters lead to misuse.

Secondly, it is not sure whether other columns will be added for identification in the future, so for scalability reasons, perhaps we can use the `label_` + `usage` format. Meaning that all columns in this category belong to `label`, this time for labeling the source, and next time add columns for labeling new uses

pkg/table-router/router.go
```diff
@@ -23,10 +24,59 @@ import (
...
-	TargetTable   string `json:"target-table" toml:"target-table" yaml:"target-table"`
+	SchemaPattern string     `json:"schema-pattern" toml:"schema-pattern" yaml:"schema-pattern"`
+	TablePattern  string     `json:"table-pattern" toml:"table-pattern" yaml:"table-pattern"`
+	TargetSchema  string     `json:"target-schema" toml:"target-schema" yaml:"target-schema"`
+	TargetTable   string     `json:"target-table" toml:"target-table" yaml:"target-table"`
+	Extractors    Extractors `json:",inline" toml:",inline" yaml:",inline"`
+}
+
+// Extractors is a rule that extracts certain values into a column
+type Extractors struct {
```
lance6716: seems if we use `omitempty`, we can keep compatibility if not enable them.

pkg/table-router/router.go
```diff
@@ -23,10 +24,59 @@ import (
...
+	TableExtractor  *TableExtractor  `json:"extract-table" toml:"extract-table" yaml:"extract-table"`
+	SchemaExtractor *SchemaExtractor `json:"extract-schema" toml:"extract-schema" yaml:"extract-schema"`
+	SourceExtractor *SourceExtractor `json:"extract-source" toml:"extract-source" yaml:"extract-source"`
+}
+
+// TableExtractor extracts table name to column
+type TableExtractor struct {
+	TargetColumn string `json:"target-column" toml:"target-column" yaml:"target-column"`
+	TableRegexp  string `json:"table-regexp" toml:"table-regexp" yaml:"table-regexp"`
+	Regexp       *regexp.Regexp
```
lance6716: maybe `Regexp` can made private because this package has fulfilled it's functionality?

pkg/table-router/router.go
```diff
@@ -23,10 +24,59 @@ import (
...
+type SourceExtractor struct {
+	TargetColumn string `json:"target-column" toml:"target-column" yaml:"target-column"`
+	SourceRegexp string `json:"source-regexp" toml:"source-regexp" yaml:"source-regexp"`
+	Regexp       *regexp.Regexp
+}
+
+// MatchVal match value via regexp
+func (t *Extractors) MatchVal(s string, ext interface{}) string {
+	var params []string
+	switch ext.(type) {
```
lance6716: maybe we can use [type switch](https://tour.golang.org/methods/16)

```
switch e := ext.(type) {
case *TableExtractor:
    params = e.Regexp.FindStringSubmatch(s)
```

pkg/table-router/router.go
```diff
@@ -39,6 +89,36 @@ func (t *TableRule) Valid() error {
 		return errors.New("target schema of table route rule should not be empty")
 	}
 
+	if t.Extractors.TableExtractor != nil {
+		re, err := regexp.Compile(t.Extractors.TableExtractor.TableRegexp)
+		if err != nil {
+			return errors.New("table extractor table regexp illegal")
```
lance6716: it's better to also display the original regexp string. and same for SchemaExtractor, SourceExtractor

pkg/table-router/router_test.go
lance6716: Hi @yufan022 it's good to add some example for new added fields in unit tests. Because for now I don't know the grammar of `TableRegexp`. Should it contain `()` to support capturing groups, or the whole regexp is to match a part of table names?

pkg/table-router/router.go
yufan022: fixed

pkg/table-router/router.go
yufan022: fixed

pkg/table-router/router.go
yufan022: fixed

pkg/table-router/router.go
lance6716: Hi @yufan022 don't forget this comment (and others). Some downgrading technique is marshaling configurations and loading the configuration files to a cluster of lower version. Currently even if we don't set `Extractors`, it will still generate extractors fields which is unrecognized by lower version DM.

yufan022: @lance6716 
```
{"schema-pattern":"core","table-pattern":"trade_*","target-schema":"core","target-table":"trade","Extractors":{}}
```
When we didn't set `TableExtractor/SchemaExtractor/SourceExtractor`, `"Extractors":{}` will also be initialized.

yufan022: @lance6716 
I will add `omitempty` on `TableExtractor/SchemaExtractor/SourceExtractor`. But `Extractors` will always be initialized because it is not a pointer. Does this cause compatibility problems?

lance6716:
ping @yufan022

yufan022:
> ping @yufan022

I left some comments to discuss.

lance6716:
> > ping @yufan022
> 
> I left some comments to discuss.

I checked the unresolved comments and now we should have agreed on your interface, as stated in https://github.com/pingcap/dm/issues/2111#issuecomment-939216053 . So for this PR I only request for add `omitempty ` and a simplt unit test to show its usage now.

pkg/table-router/router_test.go
yufan022: Got it, I will refine the unit test later.

pkg/table-router/router.go
lance6716: > Does this cause compatibility problems?

Yes. I expect `Extractors` should be totally hidden. Currently a task config exported by this version will cause "field Extractor not found in type router.TableRule" when loaded to old version.

yufan022: > I expect Extractors should be totally hidden

Seems `Extractors` can't be convert to pointer because it has `inline`, it will always initialized even if I set `,inline,omitempty`.

So, in this case, I think I have to delete the `Extractors` field. and move `TableExtractor/SchemaExtractor/SourceExtractor` to `TableRule`. 

Or do you have any good ideas?

lance6716: It's good to me that we move `TableExtractor/SchemaExtractor/SourceExtractor` to `TableRule` 👍

pkg/table-router/router.go
```diff
@@ -179,3 +238,71 @@ func (r *Table) Route(schema, table string) (string, string, error) {
...
+	for idx, param := range params {
+		if idx > 0 {
+			val += param
+		}
+	}
+	return val
+}
+
+// FetchExtendColumn get extract rule.
+func (r *Table) FetchExtendColumn(schema, table, source string) ([]string, []string) {
```
yufan022, line 263: Move the dm repository `FetchTableTransferColumnRule` here.

pkg/table-router/router.go
```diff
@@ -179,3 +238,71 @@ func (r *Table) Route(schema, table string) (string, string, error) {
...
+	var params []string
+	switch e := ext.(type) {
+	case *TableExtractor:
+		params = e.regexp.FindStringSubmatch(s)
+	case *SchemaExtractor:
+		params = e.regexp.FindStringSubmatch(s)
+	case *SourceExtractor:
+		params = e.regexp.FindStringSubmatch(s)
+	}
+	var val string
```
lance6716: we can use a https://pkg.go.dev/strings#Builder for better performance

lance6716 (commented):
rest lgtm

lance6716 (approved):

pkg/table-router/router.go
```diff
@@ -179,3 +238,71 @@ func (r *Table) Route(schema, table string) (string, string, error) {
...
+	var val strings.Builder
+	for idx, param := range params {
+		if idx > 0 {
+			val.WriteString(param)
+		}
+	}
+	return val.String()
+}
+
+// FetchExtendColumn get extract rule.
```
Ehco1996:
```suggestion
// FetchExtendColumn get extract rule, return extracted cols and extracted vals.
```

Ehco1996 (commented):
rest LGTM

Ehco1996 (approved):

yufan022:
By the way, why is `go.mod1`?

lance6716:
> By the way, why is `go.mod1`?

when using this repo as a library, we don't want itself to manage dependency versions. A historical problem

Ehco1996:
> By the way, why is `go.mod1`?

The go.mod1 files are used to break the cyclic dependency between pingcap/tidb and this module. see more here https://github.com/pingcap/tidb-tools/pull/440#issuecomment-845713951

PR #528, opened 2021-10-12
table-filter: add column filter

0xPoe:

ref: https://github.com/pingcap/ticdc/pull/3030

part of https://github.com/pingcap/ticdc/issues/3082

add column filter. The rules are the same as table filter except that there is no separator.
### Check List 

Tests 

 - Unit test

Code changes

Side effects

None

Related changes
None


0xPoe:
/cc @kennytm @lichunzhu @overvenus 

I don't know if the current extension of parse makes sense, maybe there is a better way to extend and abstract it. Please feel free to make suggestions! Thanks!

0xPoe:
/hold

Please do not merge it until TiCDC accepts the RFC for this feature. I will submit the RFC to TiCDC tomorrow.

pkg/table-filter/column_filter.go
```diff
@@ -0,0 +1,60 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package filter
+
+import "strings"
+
+// ColumnFilter is a structure to check if a column should be included for processing.
+type ColumnFilter interface {
```
3AceShowHand, line 19: how about add a interface, named `Matcher`, then implement a concret type named `ColumnSelector`

0xPoe: This is already an interface, and columnFilter is its implementation.

pkg/table-filter/helper.go
```diff
@@ -0,0 +1,23 @@
...
+package filter
+
+import "sort"
+
+// reversing replace the contents of a slice with the same elements but in reverse order.
+func reversing(items interface{}) {
+	sort.SliceStable(items, func(i, j int) bool {
+		return true
+	})
+}
```
kennytm, line 26: wut, is this an O(n log n) reverse method 😭

0xPoe: Fixed! Thanks for your review!

pkg/table-filter/column_filter.go
```diff
@@ -0,0 +1,60 @@
...
+
+	return columnFilter(p.rules), nil
+}
+
+// MatchColumn checks if a column can be processed after applying the columnFilter `f`.
+// Column is not case-sensitive on any platform, nor are column aliases.
+// So we always match in lowercase.
+func (f columnFilter) MatchColumn(column string) bool {
+	for _, rule := range f {
+		if rule.column.matchString(strings.ToLower(column)) {
```
kennytm: move the computation of `strings.ToLower(column)` out of the loop

kennytm (commented):
rest lgtm.

is the "cdc rfc" submitted?

pkg/table-filter/column_filter.go
0xPoe: Fixed. Good catch! Thanks!

0xPoe:
> is the "cdc rfc" submitted?

See: https://github.com/pingcap/ticdc/pull/3030

pkg/table-filter/column_filter_test.go
```diff
@@ -0,0 +1,305 @@
...
+		_, err := filter.ParseColumnFilter([]string{tc.arg})
+		c.Assert(err, ErrorMatches, tc.msg, Commentf("test case = %s", tc.arg))
+	}
+}
+
+func (s *columnFilterSuite) TestImport(c *C) {
+	dir := c.MkDir()
+	path1 := filepath.Join(dir, "1.txt")
+	path2 := filepath.Join(dir, "2.txt")
+	ioutil.WriteFile(path1, []byte(`
```
sleepymole: Nit: `ioutil.WriteFile` can be replaced by `os.WriteFile`.

0xPoe: Addressed! Thanks for your review!

pkg/table-filter/column_filter.go
```diff
@@ -0,0 +1,61 @@
...
+		},
+	}
+
+	for _, arg := range args {
+		if err := p.parse(arg, true); err != nil {
+			return nil, err
+		}
+	}
+
+	reversing(p.rules)
```
lichunzhu:
```suggestion
	reverse(p.rules)
```

0xPoe:
/unhold

RFC is being merged.

@kennytm @gozssky Could you please take a look? Thanks!

lichunzhu (approved):
LGTM

pkg/table-filter/column_filter.go
```diff
@@ -0,0 +1,61 @@
...
+
+	return columnFilter(p.rules), nil
+}
+
+// MatchColumn checks if a column can be processed after applying the columnFilter `f`.
+// Column is not case-sensitive on any platform, nor are column aliases.
```
lichunzhu, line 51: How about putting the reference doc here.
https://dev.mysql.com/doc/refman/5.7/en/identifier-case-sensitivity.html

pkg/table-filter/column_filter.go
0xPoe: Thanks!

pkg/table-filter/column_filter.go
0xPoe: Good idea! Added. Thanks for your review!

sleepymole (approved):

PR #529, opened 2021-10-12
*: update tidb

disksing:
Signed-off-by: disksing <i@disksing.com>

`ticdc` needs to update tidb to a newer version and `ticdc` also imports `tidb-tools`, so we have to update `tidb-tools` first.

- update tidb
- replace `pingcap/parser` with `pingcap/tidb/parser`

### Check List 

Tests 

 - Unit test


lance6716 (approved):

glorv (approved):

disksing:
The CI failure could be caused by a bug of TiDB.

Ehco1996:
> The CI failure could be caused by a bug of TiDB.

😅
have we found the root casue or is there a tracking issue ?

lance6716:
replaced by #532

PR #536, opened 2021-10-19
diff2 : fix panic

Leavrth:

1. sourceTableMaps may be empty, so `mysql_shard` will return nil when called `GetDB()`.

2. cannot handle error from `BinGenerate`

1. add check when init source

2. print the error, use origin chunk to do compare-data

### DIFF
--- a/sync_diff_inspector/utils/utils.go
+++ b/sync_diff_inspector/utils/utils.go
@@ -603,6 +603,7 @@ func GetApproximateMidBySize(ctx context.Context, db *sql.DB, schema, table stri
 		limitRange,
 		strings.Join(columnNames, ", "),
 		strconv.FormatInt(count/2, 10))
+	log.Debug("get mid by size", zap.String("sql", query), zap.Reflect("args", args))
 	rows, err := db.QueryContext(ctx, query, args...)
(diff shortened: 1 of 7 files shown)


sync_diff_inspector/utils/utils.go
```diff
@@ -582,19 +582,19 @@ func SliceToMap(slice []string) map[string]interface{} {
...
 		+------+---------+-----------------+
 		1 row in set (0.09 sec)
 	*/
-	columnNames := make([]string, 0, len(tbInfo.Columns))
-	for _, col := range tbInfo.Columns {
+	columnNames := make([]string, 0, len(indexColumns))
+	for _, col := range indexColumns {
 		columnNames = append(columnNames, dbutil.ColumnName(col.Name.O))
 	}
 	query := fmt.Sprintf("SELECT %s FROM %s WHERE %s ORDER BY %s LIMIT %s,1",
```
3pointer:
```suggestion
	query := fmt.Sprintf("SELECT %s FROM %s WHERE %s ORDER BY %s LIMIT %d,1",
```

3pointer (approved):

PR #543, opened 2021-11-04
sync_diff_inspector: Unified time zone

Leavrth:

downstream don't initialize the specific time zone when the connection is created.

use the same time_zone parameter when the upstream/downstream's connection is created.

### DIFF
--- a/sync_diff_inspector/source/source.go
+++ b/sync_diff_inspector/source/source.go
@@ -184,20 +184,17 @@ func buildSourceFromCfg(ctx context.Context, tableDiffs []*common.TableDiff, che
 }
 
 func initDBConn(ctx context.Context, cfg *config.Config) error {
+	// Unified time zone
+	vars := map[string]string{
+		"time_zone": "+0:00",
+	}
 	// we had 3 producers and `cfg.CheckThreadCount` consumer to use db connections.
 	// so the connection count need to be cfg.CheckThreadCount + 3.
-	targetConn, err := common.CreateDB(ctx, cfg.Task.TargetInstance.ToDBConfig(), nil, cfg.CheckThreadCount+3)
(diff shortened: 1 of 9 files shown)


sync_diff_inspector/utils/utils.go
```diff
@@ -723,6 +726,7 @@ func ResetColumns(tableInfo *model.TableInfo, columns []string) *model.TableInfo
 	for i, col := range tableInfo.Columns {
 		col.Offset = i
 		colMap[col.Name.O] = i
+		hasTimeStampType = hasTimeStampType || (col.FieldType.Tp == mysql.TypeTimestamp)
```
3pointer, line 729: Should we handle TypeDatetime also?

Leavrth: The time zone does not affect datetime.

3pointer (approved):

sync_diff_inspector/source/source.go
```diff
@@ -115,16 +117,18 @@ func NewSources(ctx context.Context, cfg *config.Config) (downstream Source, ups
 	tableDiffs := make([]*common.TableDiff, 0, len(tablesToBeCheck))
 	for _, tables := range tablesToBeCheck {
 		for _, tableConfig := range tables {
+			new_Info, needUnifiedTimeZone := utils.ResetColumns(tableConfig.TargetTableInfo, tableConfig.IgnoreColumns)
```
lance6716:
```suggestion
			newInfo, needUnifiedTimeZone := utils.ResetColumns(tableConfig.TargetTableInfo, tableConfig.IgnoreColumns)
```

lance6716 (commented):
rest lgtm

lance6716 (approved):

PR #547, opened 2021-11-12
sync-diff-inspector: Clear some codes and optimize GetRandomValues

Leavrth:
Issue Number: ref #670 

1. There are some fields and functions not used.
2. dbutil.GetRandomValues only get a column each time, we can get the entire rows instead.

1. remove them.
2. optimize.


sync_diff_inspector/config/config.go
```diff
@@ -180,12 +180,8 @@ func (t *TaskConfig) Init(
 		return errors.Trace(err)
 	}
 
-	// Set default value when output is empty
 	if t.OutputDir == "" {
-		t.OutputDir = timestampOutputDir()
-		if err := os.RemoveAll(t.OutputDir); err != nil && !os.IsNotExist(err) {
-			log.Fatal("fail to remove the temp directory", zap.String("path", t.OutputDir), zap.String("error", err.Error()))
-		}
+		return errors.New("output-dir is missing from the task configuration")
```
3pointer: 🤔 seems we will generate a temporary output-dir

Leavrth: PTAL again

3pointer (approved):
LGTM

lichunzhu (approved):
LGTM

PR #552, opened 2021-11-19
sync-diff-inspector: mark host for upstream/downstream in the log

Leavrth:

the log only shows that  the error occurs from the upstream or the downstream without the specific host.

mark host for upstream/downstream in the log.


3pointer (commented):
LGTM

Leavrth:
too much conflicts with the branch `master`.

PR #553, opened 2021-11-23
sync-diff-inspector: add default outputDir and support to export the template config

### LINKED ISSUE #570: outputDir conflict if using the default dir twice
## Bug Report

1. What did you do?

- Run twice with the same command: ./sync_diff_inspector --config ~/new-config.toml
- It's not user friendly that outputDir is hard coded.

3. What did you see instead?
# ./sync_diff_inspector --config ~/new-config.toml
Fail to initialize config.
failed to init Task: config changes breaking the checkpoint, please use another outputDir and start over again!

5. which tool are you using?
  sync diff 5.3.0

Leavrth:

Issue Number: close #570

1. It's not convenient to test for different config quickly.
2. no config template to refer in local.

1. add default outputDir
```
        '/tmp/sync-diff.output.2021-11-23T11.15.28+0800/fix-on-tidb0/'
You can view the comparision details through '/tmp/sync-diff.output.2021-11-23T11.15.28+0800/sync_diff.log'
```
2. support to export the template config in the current directory
```
#  ./sync-diff-inspector --help
Usage of diff:
  -V, --version                  print version of sync_diff_inspector
  -L, --log-level string         log level: debug, info, warn, error, fatal (default "info")
  -C, --config string            Config file
  -T, --template string          <dm|norm> export a template config file in the current directory
      --dm-addr string           the address of DM
      --dm-task string           identifier of dm task
      --check-thread-count int   how many goroutines are created to check data (default 1)
      --export-fix-sql           set true if want to compare rows or set to false will only compare checksum (default true)
      --check-struct-only        ignore check table's data
#  ./sync-diff-inspector -T dm
#  ./sync-diff-inspector -T norm
#  ./sync-diff-inspector -T a
Error: unexpect template name: a
-T dm: export a dm config
-T norm: export a normal config

```


3pointer (approved):
LGTM

PR #557, opened 2021-11-26
sync-diff-inspector: fix report error panic

Leavrth:

1. panic when print report error
2. get table size error should not be a table error.

add check, and remove the error trigger in `GetTableSize`


sync_diff_inspector/report/report_test.go
```diff
@@ -247,12 +253,14 @@ func TestPrint(t *testing.T) {
 		"You can view the comparision details through 'output_dir/sync_diff.log'\n")
 
 	// Error
+	report.SetTableMeetError("test", "tbl1", errors.New("123"))
+	report.SetTableStructCheckResult("test", "tbl1", false, false)
 	report.SetTableMeetError("test", "tbl", errors.New("123"))
```
3pointer: `tbl` should not meet error

3pointer (commented):
rest LGTM

3pointer (approved):

sync_diff_inspector/report/report_test.go
```diff
@@ -247,12 +253,12 @@ func TestPrint(t *testing.T) {
 		"You can view the comparision details through 'output_dir/sync_diff.log'\n")
 
 	// Error
-	report.SetTableMeetError("test", "tbl", errors.New("123"))
-	report.SetTableStructCheckResult("test", "tbl", false, false)
+	report.SetTableMeetError("test", "tbl1", errors.New("123"))
```
lance6716, line 256: we should use `nil` rather than error to test "panic when print report error"?

3pointer: this case is used to test this [line](https://github.com/pingcap/tidb-tools/pull/557/files#diff-29a3aeea519356e5b5454660011a72ebcb8508d1252d11696c23f15231110a87R241)🤔，before this PR it will panic.

lance6716 (approved):

Issue #558, opened 2021-11-28
sync-diff-inspector: mysql to mysql check error


teckick:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
If possible, provide a recipe for reproducing the error.

Create two databases and tables in one MySQL instance.

```
CREATE DATABASE inspector_mysql_0;
CREATE TABLE inspector_mysql_0.tb_emp1 (id INT(11), name VARCHAR(25), deptId INT(11));
INSERT INTO inspector_mysql_0.tb_emp1 values (1,"hello1",1);
CREATE DATABASE inspector_mysql_1;
CREATE TABLE inspector_mysql_1.tb_emp1 (id INT(11), name VARCHAR(25), deptId INT(11));
INSERT INTO inspector_mysql_1.tb_emp1 values (1,"hello1",1);
```

then run sync-diff-inspector with config.toml

```toml
# Diff Configuration.

######################### Global config #########################

# how many goroutines are created to check data
check-thread-count = 4

# set false if just want compare data by checksum, will skip select data when checksum is not equal.
# set true if want compare all different rows, will slow down the total compare time.
export-fix-sql = true

# ignore check table's data
check-struct-only = false


######################### Databases config #########################
[data-sources.mysql1]
    host = "127.0.0.1"
    port = 3306
    user = "root"
    password = "12344321"
    route-rules = ["rule1"]

[data-sources.tidb0]
    host = "127.0.0.1"
    port = 3306
    user = "root"
    password = "12344321"
    # remove comment if use tidb's snapshot data

[routes.rule1]
schema-pattern = "inspector_mysql_0"      # schema to match. Support wildcard characters * and ?.
table-pattern = "tb_emp1"          # table to match. Support wildcard characters * and ?.
target-schema = "inspector_mysql_1"         # target schema
target-table = "tb_emp1"             # target table

######################### Task config #########################
# Required
[task]
    # 1 fix sql: fix-target-TIDB1.sql
    # 2 log: sync-diff.log
    # 3 summary: summary.txt
    # 4 checkpoint: a dir
    output-dir = "/tmp/output/config"

    source-instances = ["mysql1"]

    target-instance = "tidb0"

    # tables need to check. *Include `schema` and `table`. Use `.` to split*
    target-check-tables = ["inspector_mysql_1.tb_emp1"]

    # extra table config
#     target-configs= []
    target-configs= ["config1"]

# Optional
# [table-configs]
[table-configs.config1]

[...]

PR #559, opened 2021-11-28
sync-diff-inspector: fix mysql source

teckick:

solve #558

### Check List 

Tests 

start MySQL Instance locally, and then prepare data with script in #558 

Code changes

Side effects

Related changes


sync_diff_inspector/source/source.go
```diff
@@ -130,19 +129,6 @@ func NewSources(ctx context.Context, cfg *config.Config) (downstream Source, ups
 			Collation:           tableConfig.Collation,
 			ChunkSize:           tableConfig.ChunkSize,
 		})
-
```
3pointer, line 133: we do have another requirement that ignore insensitive table name with sync-diff.
so we add the default rule to make sure the same name between upstream and downstream can be handle correctly.

@leoppro WDYT?

Leavrth:
solved by another way, details in https://github.com/pingcap/tidb-tools/issues/558

PR #577, opened 2021-12-28
sync-diff: fix panic in meetError

3pointer:

When synd-diff enter Error state. not every tables has the error. so we should filter the table doesn't have the error and only write errors into summary.
1.  filter the table doesn't have the error and only write errors into summary.

### Check List 

Tests 

 - Unit test


3pointer:
dup of #557

Issue #609, opened 2022-04-02
sync-diff-inspector:filter sequence


cartersz:
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. What did you do?
when get tables from schema, "GetTables" method in common.go do not filter sequence

2. What did you expect to see?
it can filter sequence

3. What did you see instead?

4. What version of TiDB are you using (`tidb-server -V` or run `select tidb_version();` on TiDB)?

5. which tool are you using?

6. what versionof tool are you using (`pump -V` or `tidb-lightning -V` or `syncer -V`)?

kennytm:
fixed since v5.3.2 in #610.

PR #610, opened 2022-04-02
filter sequence

cartersz:

Issue Number: ref #609 

when get tables from schema, it should filter sequence

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### DIFF
--- a/pkg/dbutil/common.go
+++ b/pkg/dbutil/common.go
@@ -375,7 +375,7 @@ func GetTables(ctx context.Context, db QueryExecutor, schemaName string) (tables
 		| NTEST          | BASE TABLE |
 		+----------------+------------+
 	*/
-	query := fmt.Sprintf("SHOW FULL TABLES IN `%s` WHERE Table_Type != 'VIEW' and Table_Type != 'SEQUENCE';", escapeName(schemaName))
+	query := fmt.Sprintf("SHOW FULL TABLES IN `%s` WHERE Table_Type = 'BASE TABLE';", escapeName(schemaName))
 	return queryTables(ctx, db, query)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/dbutil/common.go
```diff
@@ -375,7 +375,7 @@ func GetTables(ctx context.Context, db QueryExecutor, schemaName string) (tables
 		| NTEST          | BASE TABLE |
 		+----------------+------------+
 	*/
-	query := fmt.Sprintf("SHOW FULL TABLES IN `%s` WHERE Table_Type != 'VIEW';", escapeName(schemaName))
+	query := fmt.Sprintf("SHOW FULL TABLES IN `%s` WHERE Table_Type != 'VIEW' and Table_Type != 'SEQUENCE';", escapeName(schemaName))
```
lichunzhu: What about
```sql
	query := fmt.Sprintf("SHOW FULL TABLES IN `%s` WHERE Table_Type = 'BASE TABLE';", escapeName(schemaName))
```

lichunzhu: What about
```sql
	query := fmt.Sprintf("SHOW FULL TABLES IN `%s` WHERE Table_Type = 'BASE TABLE';", escapeName(schemaName))
```

cartersz: if "BASE TABLE" can include all user defined tables, i think it ok.

cartersz: @lichunzhu done

cartersz: @lichunzhu PTAL

lichunzhu (approved):
LGTM

pkg/dbutil/common.go
cartersz: lance6716 PTAL

Ehco1996 (approved):
LGTM

PR #612, opened 2022-04-25
sync-diff-inspector: let user router rule cover the default rule

### LINKED ISSUE #558: sync-diff-inspector: mysql to mysql check error

Leavrth:

Issue Number: close #558 

The function [`router.Route`](https://github.com/pingcap/tidb-tools/blob/3f5c34ef7e62e63b3a3410fe2d77bcfd9bea2934/pkg/table-router/router.go#L185) returns itself when there is no matching, So even though the rule has been replaced or removed, the target table will be still matched by the source upstream. 

So we need to add a sheild rule, such as `target-schema`.`target-table` -> `_____`.`_____`, for the source if needed.

### Check List 

Tests 

 - Unit test
 - Integration test


sync_diff_inspector/source/source.go
```diff
@@ -134,23 +134,48 @@ func NewSources(ctx context.Context, cfg *config.Config) (downstream Source, ups
...
+				// There is a user rule routing to `tableConfig.Schema`.`tableConfig.Table`
+				rules := d.Router.Match(tableConfig.Schema, tableConfig.Table)
+
+				if len(rules) == 0 {
+					// There is no self match in these user rules.
+					// Need to shield the table for this source.
+					if d.Router.AddRule(&router.TableRule{
+						SchemaPattern: tableConfig.Schema,
+						TablePattern:  tableConfig.Table,
+						TargetSchema:  "_____",
```
3pointer: maybe use other stranger name and make it const. user may have the table name like '_____'?

3pointer: image

3pointer:
rest LGTM

sync_diff_inspector/source/source.go
Leavrth: done

PR #624, opened 2022-06-06
sync-diff-inspector: fix index fields not taking effect

### LINKED ISSUE #623: sync-diff-inspector: index-fields does not take effect if bucket stats are used.
## Bug Report

1. What did you do?
```
root@sbtest[v5]> show create table sbtest1 \G
*************************** 1. row ***************************
       Table: sbtest1
Create Table: CREATE TABLE `sbtest1` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `k` int(11) NOT NULL DEFAULT '0',
  `c` char(120) NOT NULL DEFAULT '',
  `pad` char(60) NOT NULL DEFAULT '',
  PRIMARY KEY (`id`) /*T![clustered_index] CLUSTERED */,
  KEY `k_1` (`k`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin AUTO_INCREMENT=10407284
1 row in set (0.00 sec)
```

```toml
# Diff Configuration.
check-thread-count = 4
export-fix-sql = true
check-struct-only = false


######################### Datasource config #########################
[data-sources]
[data-sources.upstream]
    host = "upstream-host"
    port = 4000
    user = "root"
    password = ""
    #snapshot = "433361056758235137"

[data-sources.downstream]
    host = "downstream-host"
    port = 5300
    user = "root"
    password = ""
    #snapshot = "433361057203093509"

######################### Task config #########################
[task]
    output-dir = "./output5"
    source-instances = ["upstream"]
    target-instance = "downstream"
    target-check-tables = ["sbtest.sbtest1"]
    target-configs = ["config1"]

######################### Table config #########################
[table-configs.config1]
target-tables = ["sbtest.sbtest1"]
#range = "age > 10 AND age < 20"
index-fields = ["k"]
#ignore-columns = [""]
chunk-size = 50000
#collation = ""

```

- The key `k_1` should be used to split chunks.

3. What did you see instead?
- The primary key is used instead.

Irrelevant.

6. which tool are you using?
sync-diff-inspector

7. what versionof tool are you using (`pump -V` or `tidb-lightning -V` or `syncer -V`)?
`v5.4.1`.

liuzix:

Issue Number: close #623 

- Added `index-fields` index selection for `BucketIterator`

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test

Related changes

### DIFF
--- a/sync_diff_inspector/splitter/index_fields.go
+++ b/sync_diff_inspector/splitter/index_fields.go
@@ -1,0 +1,111 @@
+// Copyright 2022 PingCAP, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package splitter
+
+import (
+	"sort"
+	"strings"
+
+	"github.com/pingcap/errors"
+	"github.com/pingcap/log"
+	"github.com/pingcap/tidb-tools/sync_diff_inspector/utils"
+	"github.com/pingcap/tidb/parser/model"
+	"go.uber.org/zap"
+)
+
+// indexFields wraps the column info for the user config "index-fields".
+type indexFields struct {
+	cols      []*model.ColumnInfo
+	tableInfo *model.TableInfo
+	empty     bool
+}
+
+func indexFieldsFromConfigString(strFields string, tableInfo *model.TableInfo) (*indexFields, error) {
+	if len(strFields) == 0 {
+		// Empty option
+		return &indexFields{empty: true}, nil
+	}
+
+	if tableInfo == nil {
+		log.Panic("parsing index fields with empty tableInfo",
+			zap.String("index-fields", strFields))
+	}
+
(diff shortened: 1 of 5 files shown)


sync_diff_inspector/splitter/index_fields.go
```diff
@@ -0,0 +1,111 @@
...
+		log.Panic("matching with empty index")
+	}
+	if len(f.cols) == 0 {
+		log.Panic("unexpected cols with length 0")
+	}
+
+	if len(index.Columns) != len(f.cols) {
+		// We need an exact match.
+		// Lengths not matching eliminates the possibility.
+		return false
```
lichunzhu, line 81: If the given columns can match the first few columns, this index can still work?

liuzix: The real intention of the user may be to use columns that are part of a composite key. If we allow matching here, it does not do what the user expects.

lichunzhu: If composite key is `(a,b,c)` and users want to use columns `a,b`, I think we can still match this index? However, `b,a` or `b,a,c` can't be matched.

3pointer:
I feel like the real problem is we didn't handle `limit` in `bucketSplitter`.  so maybe we should fallback to `randomSplitter` when user set the `range` config. and also we need this PR to handle index-fields properly

3pointer (approved):
LGTM

lichunzhu (approved):
LGTM

PR #628, opened 2022-06-13
dbutil: add `interpolateParams=true` into DSN

### LINKED ISSUE #627: dbutil: sql query inconsistence
## Bug Report

1. What did you do?
Using string variable as arguments of query may lead to result inconsistence.
For example, there is a query : `SELECT .... WHERE id > ?;`
We have 2 kinds of args:
```
args1 := []interface{}{"4611686018427432302"}
args2 := []interface{}{4611686018427432302}
```

And we do that query:
```
conn.Query(query, args1...)...
conn.Qeury(qeury, args2...)...
```

the result is the same as that using long int variable.

3. What did you see instead?
they are different.

5. which tool are you using?

Leavrth:

Issue Number: close #627 

add `interpolateParams=true` into DSN

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code


lichunzhu (approved):

3pointer (approved):
LGTM

PR #630, opened 2022-06-14
sync_diff_inspector: add config item: chunk-thread-count

### LINKED ISSUE #629: sync-diff-inspector has a slow speed of splitting chunks in the case of multiple tables
## Bug Report

sync-diff-inspector has only 3 fixed number of goroutines to split chunks. When there are many tables, this becomes a bottleneck.

1. What did you do?
use sync-diff-inspector to check many small tables.

it should be quickly.

3. What did you see instead?
it is slow, so we need to run multiple sync-diff-inspectors at the same time to speed up.

5. which tool are you using?
sync-diff-inspector

Leavrth:

Issue Number: close #629 

add config item `chunk-thread-count` to adjust the number of goroutines to split chunks.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes


lichunzhu (approved):

sync_diff_inspector/source/chunks_iter.go
```diff
@@ -31,31 +31,33 @@ type ChunksIterator struct {
...
 	iter := &ChunksIterator{
-		tableAnalyzer: analyzer,
-		TableDiffs:    tableDiffs,
-		chunksCh:      make(chan *splitter.RangeInfo, 64),
-		errCh:         make(chan error, len(tableDiffs)),
-		cancel:        cancel,
+		splitThreadCount: splitThreadCount,
+		tableAnalyzer:    analyzer,
+		TableDiffs:       tableDiffs,
+		chunksCh:         make(chan *splitter.RangeInfo, 30*splitThreadCount),
```
3pointer: could you add some comments about magic number 30?

Leavrth: done

Leavrth:
Previously, the sync-diff-inspector allocates a WorkerPool with `check-thread-count` goroutines for the comparison process, and split `spliter-thread-count` tables at the same time. When a table uses buckets to generate chunks, a WorkerPool with `check-count-thread` goroutines is allocated to each bucket splitter. So sync-diff-inspector needs about `check-thread-count` + `check-thread-count` * `split-thread-count` connections at most (compare process + bucket-spliter for each table).

Now, bucket-spliters share workpool, that is, sync-diff-inspector needs about `check-thread-count` + `check-thread-count` + `split-thread-count` connections at most.

3pointer:
> `check-thread-count` * `split-thread-count`

can we unified the `check-thread-count` and `split-thread-count` to one config? or maybe calculate `split-thread-count` from some exists configs.

Leavrth:
now, make `split-thread-count` unexposed, and let it equals `3+check-thread-count/2`.

3pointer (approved):

sync_diff_inspector/splitter/bucket.go
```diff
@@ -221,12 +222,13 @@ func (s *BucketIterator) splitChunkForBucket(ctx context.Context, firstBucketID,
 		chunk.InitChunks(chunks, chunk.Bucket, firstBucketID, lastBucketID, beginIndex, s.table.Collation, s.table.Range, bucketChunkCnt)
 		progress.UpdateTotal(s.progressID, len(chunks), false)
 		s.chunksCh <- chunks
+		s.wg.Done()
```
lichunzhu: if err != nil it seems that s.wg won't Done?
use this as a defer before L214?

lichunzhu (commented):
rest LGTM

lichunzhu (approved):
LGTM

PR #671, opened 2022-08-16
sync-diff-inspector: change bit_xor to sum

### LINKED ISSUE #634: BIT_ XOR function has a collision example that is easy to occur
## Bug Report

1. What did you do?
0F8r6pG7TV
`BIT_XOR` function after `CRC32` function is hard to avoid this example.
S4MaUxDP5X

The 2 tables' checksums should be different.

3. What did you see instead?
The 2 tables' checksums are the same.

5. which tool are you using?
sync-diff-inspector

```
Release Version: v5.2.2-142-g2d0c30a-dirty
Git Commit Hash: 2d0c30aae7cbdb15ab777c934dcdbcb786ea1d3a
Git Branch: config_chunk_iter_thread
UTC Build Time: 2022-06-15 07:29:11
Go Version: go1.18.3
```

Leavrth:

Issue Number: close #634 

use `SUM` instead of `BIT_XOR`.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

### DIFF
--- a/sync_diff_inspector/source/common/table_diff.go
+++ b/sync_diff_inspector/source/common/table_diff.go
@@ -62,6 +62,9 @@ type TableDiff struct {
 	// the table has column timestamp, which need to reset time_zone.
 	NeedUnifiedTimeZone bool `json:"-"`
 
+	// the table has unique column, so it's ok to use bit_xor function.
+	HasUniqueColumn bool `json:"-"`
+
 	Collation string `json:"collation"`
 
 	ChunkSize int64 `json:"chunk-size"`
--- a/sync_diff_inspector/source/mysql_shard.go
+++ b/sync_diff_inspector/source/mysql_shard.go
@@ -103,7 +103,7 @@ func (s *MySQLSources) GetCountAndCrc32(ctx context.Context, tableRange *splitte
 
 	for _, ms := range matchSources {
 		go func(ms *common.TableShardSource) {
-			count, checksum, err := utils.GetCountAndCRC32Checksum(ctx, ms.DBConn, ms.OriginSchema, ms.OriginTable, table.Info, chunk.Where, chunk.Args)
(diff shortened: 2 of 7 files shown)


GMHDBJD:
Don't think SUM is better then XOR. 🤔

Leavrth:
> Don't think SUM is better then XOR. 🤔

Yes, I think so. But `BIT_XOR` is easiler to lead to hash collision when the table has no unique column. 
I think the difference caused by special failure is easiler to appear than that by the failure such as some bits randomly changed in the data migration. Especially duplicate same rows(a xor a xor a=a, we use count to solve it), a column of some rows is changed in a regular way(#634), ...

So, how about using `SUM` function **only** for the table without unique column? @GMHDBJD

lance6716:
This is not related to UNIQUE column?

Leavrth:
> This is not related to UNIQUE column?

It seems not as simple as I thought. :hot_face:

Leavrth:
Change the content of the `concat_ws`, the results of `bit_xor(crc32(...))` of 2 rows from 2 tables are always the same.

```mysql
mysql> select BIT_XOR(cast(crc32(concat_ws('.', j, i, x, j,x, j)) as unsigned)) as CHECKSUM from db_test.t4;
+------------+
| CHECKSUM   |
+------------+
| 2478727481 |
+------------+
1 row in set (0.01 sec)

mysql> select BIT_XOR(cast(crc32(concat_ws('.', j, i, x, j,x, j)) as unsigned)) as CHECKSUM from db_test.t5;
+------------+
| CHECKSUM   |
+------------+
| 2478727481 |
+------------+
1 row in set (0.00 sec)

mysql> select * from db_test.t5;
+------+-------+------+
| i    | j     | x    |
+------+-------+------+
|    1 | 10001 |    3 |
|    1 | 10011 |    3 |
+------+-------+------+
2 rows in set (0.00 sec)

mysql> select * from db_test.t4;
+------+-------+------+
| i    | j     | x    |
+------+-------+------+
|    1 | 10001 |    2 |
|    1 | 10011 |    2 |
+------+-------+------+
2 rows in set (0.01 sec)
```

Seems that `bit_xor + crc32` is not a good idea. 🤔

PR #684, opened 2022-10-25
implement hiding password

### LINKED ISSUE #683: password is shown as plain text at sync-diff-inspector
## Feature Request

Not related to a problem

sync-diff-inspector log shows data source password as plain text, I want them to be masked because they are sensitive value for us

**Describe alternatives you've considered:**
* copy values and replace with masking here > https://github.com/pingcap/tidb-tools/blob/master/sync_diff_inspector/main.go#L101

**Teachability, Documentation, Adoption, Migration Strategy:**

YuJuncen:
Signed-off-by: hillium <yujuncen@pingcap.com>

Issue Number: close #683

This PR redacts the sensitive values in the `datasource` configure.

### Check List 

Tests 

 - Unit test
image


lance6716 (approved):

sync_diff_inspector/utils/utils.go
```diff
@@ -35,6 +35,18 @@ import (
...
+// SecretString is a wrapper for sensitive strings like password,
+// which yields redacted string when being marshaled.
+type SecretString string
+
+func (s SecretString) MarshalJSON() ([]byte, error) {
+	return []byte(`"******"`), nil
+}
+
+func (s SecretString) String() string {
+	return "******"
```
Leavrth, line 47: we need to use it to generate DSN, is it ok to use `******`?
https://github.com/pingcap/tidb-tools/blob/422258e0d5cd216d3dafb6670dee0c93be269120/sync_diff_inspector/config/config.go#L176

YuJuncen: I think perhaps we'd better manually unwrap the secret string here 🤔

Leavrth (approved):

PR #695, opened 2023-01-06
sync_diff_inspector: fix the timezone difference

### LINKED ISSUE #694: sync_diff_inspector: time_zone is difference when parse the default value of type `timestamp`
## Bug Report

1. What did you do?
+ In UTC +8:00
+ create table with a column whose type is timestamp and default value is `1970-01-01 08:00:01`.
+ use sync_diff_inspector to check tables.

4. What did you expect to see?
success to check

5. What did you see instead?
failed to check. 
```
get table test.t1's information error [types:1067]Invalid default value for 'ts'
```

6. What version of TiDB are you using (`tidb-server -V` or run `select tidb_version();` on TiDB)?

7. which tool are you using?

8. what versionof tool are you using (`pump -V` or `tidb-lightning -V` or `syncer -V`)?

Leavrth:
Signed-off-by: Leavrth <jianjun.liao@outlook.com>

Issue Number: close #694 

Because sync_diff_inspector currently set time_zone to UTC +0:00 as default, it is necessary to use an sessionCtx with UTC time_zone to parse the table info.

### Check List 

Tests 

 - No code

### DIFF
--- a/pkg/dbutil/table.go
+++ b/pkg/dbutil/table.go
@@ -18,17 +18,21 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/coreos/go-semver/semver"
 	"github.com/pingcap/errors"
+	"github.com/pingcap/parser/mysql"
 	"github.com/pingcap/tidb/ddl"
 	"github.com/pingcap/tidb/parser"
 	"github.com/pingcap/tidb/parser/ast"
 	"github.com/pingcap/tidb/parser/model"
 	_ "github.com/pingcap/tidb/planner/core" // to setup expression.EvalAstExpr. See: https://github.com/pingcap/tidb/blob/a94cff903cd1e7f3b050db782da84273ef5592f4/planner/core/optimizer.go#L202
+	"github.com/pingcap/tidb/sessionctx"
 	"github.com/pingcap/tidb/types"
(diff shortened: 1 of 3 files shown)


pkg/dbutil/table.go
```diff
@@ -110,7 +114,10 @@ func GetTableInfoWithVersion(ctx context.Context, db QueryExecutor, schemaName s
 	if err != nil {
 		return nil, errors.Trace(err)
 	}
-	return GetTableInfoBySQL(createTableSQL, parser2)
+	sctx := mock.NewContext()
+	// unify the timezone to UTC +0:00
+	sctx.GetSessionVars().TimeZone = time.UTC
```
lance6716, line 119: we should use the same time zone of `createTableSQL`? which is the timezone of SQL connection

Leavrth: we have already set session timezone to `+0:00`, so the `createTableSQL` gotten from session query is also base on `+0:00`. Besides the function `GetTableInfoWithVersion` is called only by sync_diff_inspector.

https://github.com/pingcap/tidb-tools/blob/98f55c7bc75d1521bf634ae16ccfdf1ed9ae22b1/sync_diff_inspector/config/config.go#L186

lance6716 (approved):

PR #700, opened 2023-02-18
sync_diff_inspector: fix return err when float/double is zero

### LINKED ISSUE #699: check float will return error.When float data has 0 value on source or on target
## Bug Report

1. What did you do?
source tidb
```
mysql> create table f(id int primary key,f float);
mysql> insert into f values(1,2),(2,4.5),(3,0);
```
target tidb
```
mysql> create table f(id int primary key,f float);
mysql> insert into f values(1,0),(2,0),(3,0);
```
no error

3. What did you see instead?

convert , 4.5 to float failed, err1: strconv.ParseFloat: parsing "": invalid syntax, err2: <nil> error occured in `a`.`f`

v6.5.0
5. which tool are you using?
sync_diff_inspector

erwadba:

Issue Number: close #699

When CompareData func compare real rows data. It's no need do anything about float/double type.

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

### DIFF
--- a/sync_diff_inspector/utils/utils.go
+++ b/sync_diff_inspector/utils/utils.go
@@ -543,9 +543,6 @@ func CompareData(map1, map2 map[string]*dbutil.ColumnData, orderKeyCols, columns
 				if math.Abs(num1-num2) <= 1e-6 {
 					continue
 				}
-			} else {
-				equal = false
-				break
 			}
 		} else if column.FieldType.GetType() == mysql.TypeJSON {
 			if (str1 == str2) || (data1.IsNull && data2.IsNull) {
--- a/sync_diff_inspector/utils/utils.go
+++ b/sync_diff_inspector/utils/utils.go
@@ -143,15 +143,7 @@ func GetTableRowsQueryFormat(schema, table string, tableInfo *model.TableInfo, c
 
 	columnNames := make([]string, 0, len(tableInfo.Columns))
 	for _, col := range tableInfo.Columns {
-		name := dbutil.ColumnName(col.Name.O)
-		// When col value is 0, the result is NULL.
-		// But we can use ISNULL to distinguish between null and 0.
-		if col.FieldType.GetType() == mysql.TypeFloat {
-			name = fmt.Sprintf("round(%s, 5-floor(log10(abs(%s)))) as %s", name, name, name)
-		} else if col.FieldType.GetType() == mysql.TypeDouble {
-			name = fmt.Sprintf("round(%s, 14-floor(log10(abs(%s)))) as %s", name, name, name)
-		}
-		columnNames = append(columnNames, name)
+		columnNames = append(columnNames, dbutil.ColumnName(col.Name.O))
 	}
 	columns := strings.Join(columnNames, ", ")
 	if collation != "" {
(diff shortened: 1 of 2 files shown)


erwadba:
@lance6716  @Leavrth  PTAL,tks.

sync_diff_inspector/utils/utils.go
```diff
@@ -539,18 +531,18 @@ func CompareData(map1, map2 map[string]*dbutil.ColumnData, orderKeyCols, columns
...
+			if data1.IsNull && data2.IsNull {
 				continue
+			} else if !data1.IsNull && !data2.IsNull {
+				num1, err1 := strconv.ParseFloat(str1, 64)
+				num2, err2 := strconv.ParseFloat(str2, 64)
+				if err1 != nil || err2 != nil {
+					err = errors.Errorf("convert %s, %s to float failed, err1: %v, err2: %v", str1, str2, err1, err2)
+					return
+				}
+				if math.Abs(num1-num2) <= 1e-6 {
```
lance6716, line 543: this hard coded threshold is not as good as 
```
		if col.FieldType.GetType() == mysql.TypeFloat {
			name = fmt.Sprintf("round(%s, 5-floor(log10(abs(%s)))) as %s", name, name, name)
		} else if col.FieldType.GetType() == mysql.TypeDouble {
			name = fmt.Sprintf("round(%s, 14-floor(log10(abs(%s)))) as %s", name, name, name)
		}
```

lance6716 (commented):
rest lgtm, good work

@Leavrth

sync_diff_inspector/utils/utils.go
erwadba: We can use `num1=num2` to check it is equal or not. I want to know why float/double type need to be converted when comparing. Can you give me some examples?

When I canceled this code. It can also distinguish between source and target about float/double type.
https://github.com/pingcap/tidb-tools/blob/2218c401d5d7c0541d77fbc39adcd33d3dff2d81/sync_diff_inspector/utils/utils.go#L149-L153
https://github.com/pingcap/tidb-tools/blob/2218c401d5d7c0541d77fbc39adcd33d3dff2d81/sync_diff_inspector/utils/utils.go#L764-L768

Leavrth: 🤨  This is a history issue. Currently this 'hack' sub-expression `round(%s, 5-floor(log10(abs(%s))))` is done by Database itself, in other words, the database(TiDB/MySQL) will return the same result of `select (float value)` as the result of the sub-expression.
I think the version is high enough to remove this codes. 

> When I canceled this code. It can also distinguish between source and target about float/double type.

We also need to check whether the result of checksum is `false negative`, which means some checksum should be the same but is different, leads to trigger row compare operation(cost hightly).

erwadba: Got it. tks.

Leavrth (approved):

Leavrth:
I found that in TiDB, if data has more than 6 significant digits, TiDB won't round into 6 significant digits.

Let's see the steps:
1. run these SQLs in MySQL and TiDB
```SQL
create table test.float1(a int, f float);
insert into test.float1 (a, f) values (1,0.1),(2,0.2),(3,0.3),(4,0.4), ..(skip).. ,(500,0.500);
insert into test.float1 (a, f) values(0, 0.2345678);
insert into test.float1 (a,f) values(-1, 2345678000);
insert into test.float1 (a,f) values(-2, 23.45678000);
```

2. check these data by running these SQLs
in TiDB
```SQL
mysql> select * from test.float1 where a = -2 or a = -1 or a = 0 or a  = 2;
+------+------------+
| a    | f          |
+------+------------+
|    0 |  0.2345678 |
|   -1 | 2345678000 |
|    2 |       0.22 |
|   -2 |   23.45678 |
+------+------------+
4 rows in set (0.00 sec)
```

in MySQL
```SQL
mysql> select * from test.float1 where a = -2 or a = -1 or a = 0 or a  = 2;
+------+------------+
| a    | f          |
+------+------------+
|    2 |        0.2 |
|    0 |   0.234568 |
|   -1 | 2345680000 |
|   -2 |    23.4568 |
+------+------------+
4 rows in set (0.00 sec)
```

3. run sync-diff-inspector
the table is equal. That's because checksum use `round(%s, 5-floor(log10(abs(%s))))`.

4. run these SQLs in TiDB:
```SQL
delete from test.float1 where a = 2;
insert into test.float1 (a, f) values (2, 0.22);
```

5. run sync-diff-inspector again
```
$ ./sync_diff_inspector --config ./config2.toml
A total of 1 tables need to be compared

Comparing the table structure of ``test`.`float1`` ... equivalent
Comparing the table data of ``test`.`float1`` ... failure
_____________________________________________________________________________
Progress [============================================================>] 100% 0/0
The data of `test`.`float1` is not equal

A total of 1 tables have been compared, 0 tables finished, 1 tables failed, 0 tables skipped.
        '/tmp/sync-diff.output.2023-03-06T15.25.51+0800/fix-on-tidb0/'
You can view the comparision details through '/tmp/sync-diff.output.2023-03-06T15.25.51+0800/sync_diff.log'
```
see the fix.sql
```SQL
-- table: test.float1
-- range in sequence: Full
DELETE FROM `test`.`float1` WHERE `a` = -2 AND `f` = 23.45678 LIMIT 1;
REPLACE INTO `test`.`float1`(`a`,`f`) VALUES (-2,23.4568);
DELETE FROM `test`.`float1` WHERE `a` = -1 AND `f` = 2345678000 LIMIT 1;
REPLACE INTO `test`.`float1`(`a`,`f`) VALUES (-1,2345680000);
REPLACE INTO `test`.`float1`(`a`,`f`) VALUES (2,0.2);
DELETE FROM `test`.`float1` WHERE `a` = 2 AND `f` = 0.22 LIMIT 1;
```
It seems that we can not depend on `a-b < 1e-6` to check difference.
So we still need to use `round` in the SQL.

PR #707, opened 2023-02-25
sync-diff-inspector: Use md5 func replace crc32 for checksum

erwadba:
### What problem does this PR solve?

Issue Number: close #703  #634

### What is changed and how it works?
Use md5 func replace crc32 for checksum.Data check is more accurate.

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

PR #784, opened 2024-05-04
Update sync-diff-inspector to perform row-level comparison when table checksums match

### LINKED ISSUE #634: BIT_ XOR function has a collision example that is easy to occur

michaelmdeng:

Current behavior is to perform row-level data comparison only when table checksums don't match *__and__* when user requests DML generation. Given #634, there are possible checksum collisions that cause sync-diff-inspector to return false negatives where it considers two different tables to be equal. In these scenarios, sync-diff-inspector will not compare rows and will mistakenly determine tables to be equal.

A checksum can only identify true negatives, ie. two tables are certainly different, and cannot determine true positives, ie. two tables are certainly equal. Thus a more desirable behavior is to use the checksum to remove the need for more expensive row-level data comparison when a simpler/faster checksum can already tell us the tables are different.

However, in the common case that the user wants to use sync-diff-inspector to generate DML statements when the tables are different, we need to perform row-level comparison anyway.

Issue Number: close #634 

Thus, we should perform row-level data comparison if user requests DML generation *__or__* if the checksums match. The only case where we shouldn't compare row data is when the user does not desire DML and if the checksums don't match (confirming tables are different w/out data check).

### Check List 

Tests 

 - Unit test
 - Integration test
 - [x] Manual test (add detailed scripts or steps below)
 - No code

Generate checksum collision

```sql
create table test_1.test(id int primary key,b blob);
insert into test_1.test values (1, x'FF00'), (2, x'00FF');
create table test_2.test(id int primary key,b blob);
insert into test_2.test values (1, x'00FF'), (2, x'FF00');
```

Confirm that current state considers tables equal. Confirm changed state considers tables different

Code changes

 - Has persistent data change

Side effects

 - [x] Possible performance regression
 - Increased code complexity

This change will run row-level data comparison in more cases than previous, which can be more resource-intensive than simply checksumming. However this is in exchange for more correctness in cases where checksum collisions occur.

Related changes


lyzx2001:
Although this pr can fix the false positive issue, the row-level comparison will cause huge performance regression, as it abandons the distributed computing in tikv.

We do not prefer to fix an issue with very low probability to happen at the cost of performance regression for most cases.

For now, we may still use this pr https://github.com/pingcap/tidb-tools/pull/707 to solve the problem.

michaelmdeng:
> Although this pr can fix the false positive issue, the row-level comparison will cause huge performance regression, as it abandons the distributed computing in tikv.
> 
> We do not prefer to fix an issue with very low probability to happen at the cost of performance regression for most cases.
> 
> For now, we may still use this pr #707 to solve the problem.

Ack, I've opened #787 as an alternative approach using the checksum improvements you reference.

michaelmdeng:
Closing in favor of #787

PR #831, opened 2024-11-27
sync_diff_inspector: Compare rows even if export-fix-sql is false

### LINKED ISSUE #830: Table data check fails if export-fix-sql is disabled for tables with complex json
## Bug Report

```sql
CREATE TABLE `t1` (
  `id` int NOT NULL,
  `j` json DEFAULT NULL,
  PRIMARY KEY (`id`)
);
INSERT INTO t1 VALUES (1, '{"aaa": 1, "aa": 1, "bb": 1}');
```

With `export-fix-sql = false` this would result in this:
```
$ ./bin/sync_diff_inspector -C sd.toml -L debug
A total of 1 tables need to be compared

Comparing the table structure of `test`.`t1` ... equivalent
Comparing the table data of `test`.`t1` ... failure
_____________________________________________________________________________
Progress [============================================================>] 100% 0/0
The data of `test`.`t1` is not equal

A total of 1 tables have been compared, 0 tables finished, 1 tables failed, 0 tables skipped.
	'output/fix-on-tidb0/'
You can view the comparison details through './output/sync_diff.log'
```

With `export-fix-sql = true` this would result in this:
```
$ ./bin/sync_diff_inspector -C sd.toml -L debug
A total of 1 tables need to be compared

Comparing the table structure of `test`.`t1` ... equivalent
Comparing the table data of `test`.`t1` ... equivalent
_____________________________________________________________________________
Progress [============================================================>] 100% 0/0
A total of 1 table have been compared and all are equal.
You can view the comparison details through './output/sync_diff.log'
```

What happens here is this:
- `compareChecksumAndGetCount()` would return `isEqual=false` as the checksum doesn't match due to representation differences for the JSON type.
- Then `if !isEqual && df.exportFixSQL {` would only call `compareRows()` if `exportFixSQL` is true.

So when `export-fix-sql` is set to true it will check row data if the checksum doesn't match, but this isn't done if this setting is set to false.

dveeden:

Issue Number: close #830 

### Check List 

Tests 

Side effects

 - Possible performance regression: Comparing rows is slower than comparing checksums

Related changes

```release-note
Comparing JSON with export-fix-sql being enabled was failing when it shouldn't
```

### DIFF
--- a/sync_diff_inspector/diff.go
+++ b/sync_diff_inspector/diff.go
@@ -437,7 +437,7 @@ func (df *Diff) consume(ctx context.Context, rangeInfo *splitter.RangeInfo) bool
 		// If an error occurs during the checksum phase, skip the data compare phase.
 		state = checkpoints.FailedState
 		df.report.SetTableMeetError(schema, table, err)
-	} else if !isEqual && df.exportFixSQL {
+	} else if !isEqual {
 		state = checkpoints.FailedState
 		// if the chunk's checksum differ, try to do binary check
 		info := rangeInfo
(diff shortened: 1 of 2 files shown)


mjonss (commented):
Is it possible to add a test covering this change?

dveeden:
> Is it possible to add a test covering this change?

I added a test in 6e4d86480367f244d6ce7b9b7a08c79e8584ff4d but I'm not sure if that runs in CI or not

mjonss (approved):
LGTM

Defined2014 (approved):

PR #845, opened 2025-01-31
Allow max-execution-time parameter for SyncDiffInspector

### LINKED ISSUE #844: Add max_execution_time parameter for sync diff inspector
## Feature Request

* We enforce cluster max_execution_time as a safety mechanism at 30s and find that sync diff inspector needs longer runtimes to avoid experiencing errors due to canceled requests.
* This patch allows for setting a different session level max_execution_time value to override the cluster global value.
* It maintains the default behavior (no session level limit) unless an additional new argument is provided to sync-diff-inspector

* Session level override for max_execution_time

**Describe alternatives you've considered:**
* No good ones came to mind and would require relaxing our reliability stance.

**Teachability, Documentation, Adoption, Migration Strategy:**
* N/A, upgrade and use argument

Filing follow up PR using the patch we run in production.

zph:
Allows passing max-execution-time to sync-diff-inspector to allow it to override a stricter cluster level max-execution-time.

The default remains the same that no max-execution-time will be set if the config file lacks this parameter. It's backwards compatible.

Issue Number: close #844

New argument for sync-diff-inspector to set session level max_execution_time value.

### Check List

Tests 

 - Unit test

Code changes

Side effects

 - New cli argument available w/ backwards compatible default

Related changes

# QUESTIONS
- [ ] I'm unsure how / where to update docs, please advise.
- [ ] I added the field check to an existing test, advise if you want it in new test
- [ ] I marked the field as `omitempty` to avoid backwards compatibility break if users are strictly parsing json output logs. Advise if you want to break existing interface of that log by outputting max-execution-time arg

### DIFF
--- a/sync_diff_inspector/config/config.go
+++ b/sync_diff_inspector/config/config.go
@@ -114,7 +114,7 @@ type DataSource struct {
 	Password         utils.SecretString `toml:"password" json:"password"`
 	SqlMode          string             `toml:"sql-mode" json:"sql-mode"`
 	Snapshot         string             `toml:"snapshot" json:"snapshot"`
-	MaxExecutionTime int                `toml:"max-execution-time" json:"max-execution-time"`
+	MaxExecutionTime int                `toml:"max-execution-time" json:"max-execution-time,omitempty"`
 
 	Security *Security `toml:"security" json:"security"`
 
--- a/sync_diff_inspector/config/config.go
+++ b/sync_diff_inspector/config/config.go
@@ -108,12 +108,13 @@ type Security struct {
 
 // DataSource represents the Source Config.
 type DataSource struct {
-	Host     string             `toml:"host" json:"host"`
-	Port     int                `toml:"port" json:"port"`
-	User     string             `toml:"user" json:"user"`
-	Password utils.SecretString `toml:"password" json:"password"`
-	SqlMode  string             `toml:"sql-mode" json:"sql-mode"`
-	Snapshot string             `toml:"snapshot" json:"snapshot"`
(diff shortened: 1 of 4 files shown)


sync_diff_inspector/config/config.go
```diff
@@ -108,12 +108,13 @@ type Security struct {
...
-	Password utils.SecretString `toml:"password" json:"password"`
-	SqlMode  string             `toml:"sql-mode" json:"sql-mode"`
-	Snapshot string             `toml:"snapshot" json:"snapshot"`
+	Host             string             `toml:"host" json:"host"`
+	Port             int                `toml:"port" json:"port"`
+	User             string             `toml:"user" json:"user"`
+	Password         utils.SecretString `toml:"password" json:"password"`
+	SqlMode          string             `toml:"sql-mode" json:"sql-mode"`
+	Snapshot         string             `toml:"snapshot" json:"snapshot"`
+	MaxExecutionTime int                `toml:"max-execution-time" json:"max-execution-time,omitempty"`
```
kennytm, line 117: we are planning to make it accept all session variables ([similar to `target-database.session` in DM](https://docs.pingcap.com/tidb/stable/task-configuration-file-full)), so it can support `max_execution_time` as well as other options e.g. [to tune the optimizer](https://docs.pingcap.com/tidb/stable/system-variables#tidb_opt_agg_push_down) etc.

EDIT: #847 was just submitted.

zph: @kennytm Ok thank you. I'll follow that PR instead.

I'll close this PR

PR #846, opened 2025-02-06
sync_diff_inspector: process special character in generated fix SQL

### LINKED ISSUE #833: some line breaks and backslashes are missing in fix.sql generated by sync-diff-inspector
## Bug Report

1. What did you do?
1、upstream database :  varchar field contains some newlines and backslashes (\\\)
2、downstream database : varchar field contains some newlines and backslashes (\)
3、then use sync-diff-inspector compare data,
4、Some line breaks and backslashes are missing in the _**fix.sql**_ 

fix.sql have all line breaks and backslashes , just like upsteams data row.

3. What did you see instead?
Some line breaks and backslashes are missing in the _**fix.sql**_ 

upstream : v3.0.8
downstream: v6.5.7
sync-diff-inspector : v6.5.3

5. which tool are you using?
sync-diff-inspector : v6.5.3

joechenrh:

Issue Number: close #833

Add missing process for characters like backslashes (\\) in fix SQL.

### Check List 

Tests 

 - Unit test
 - Integration test

Related changes

```
process special characters like backslashes in generated fix SQL.
```

### DIFF
--- a/sync_diff_inspector/utils/utils.go
+++ b/sync_diff_inspector/utils/utils.go
@@ -172,6 +172,14 @@ func GetTableRowsQueryFormat(schema, table string, tableInfo *model.TableInfo, c
 	return query, orderKeyCols
 }
 
+// escapeString escapes special characters in the given string
+func escapeString(bs []byte) string {
+	s := string(bs)
+	s = strings.Replace(s, "\\", "\\\\", -1)
+	s = strings.Replace(s, "'", "\\'", -1)
+	return s
+}
+
 // GenerateReplaceDML returns the insert SQL for the specific row values.
 func GenerateReplaceDML(data map[string]*dbutil.ColumnData, table *model.TableInfo, schema string) string {
 	colNames := make([]string, 0, len(table.Columns))
@@ -191,7 +199,7 @@ func GenerateReplaceDML(data map[string]*dbutil.ColumnData, table *model.TableIn
 			if dbutil.IsBlobType(col.FieldType.GetType()) || IsBinaryColumn(col) {
 				values = append(values, fmt.Sprintf("x'%x'", data[col.Name.O].Data))
 			} else {
-				values = append(values, fmt.Sprintf("'%s'", strings.Replace(string(data[col.Name.O].Data), "'", "\\'", -1)))
+				values = append(values, fmt.Sprintf("'%s'", escapeString(data[col.Name.O].Data)))
 			}
 		} else {
 			values = append(values, string(data[col.Name.O].Data))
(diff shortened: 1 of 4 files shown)


kennytm (approved):

sync_diff_inspector/utils/utils.go
```diff
@@ -172,6 +172,14 @@ func GetTableRowsQueryFormat(schema, table string, tableInfo *model.TableInfo, c
 	return query, orderKeyCols
 }
 
+// escapeString escapes special characters in the given string
+func escapeString(bs []byte) string {
+	s := string(bs)
+	s = strings.Replace(s, "\\", "\\\\", -1)
+	s = strings.Replace(s, "'", "\\'", -1)
+	return s
```
D3Hunter, line 180: any reference for the escape rule? do we need to handle new line? I see the issue mention it `Some line breaks and backslashes are missing in the fix.sql`

joechenrh: I've added linebreaks and backslashes in the UT:
https://github.com/pingcap/tidb-tools/pull/846/files#diff-27471957ca7f3cc0b159ef57928a0bcbb7de33a4cbe224bb48d3f81c6085da91R350-R351
And I also did some tests locally. Let me add an integration test.

kennytm: if you also need to escape `\n` and `\r` etc perhaps just `import "github.com/pingcap/tidb/util/format"` and use its `OutputFormat` function

joechenrh: > if you also need to escape \n and \r

Since the input is a byte slice, I think we don't need additional process for `\n` and `\r`?

D3Hunter: > Since the input is a byte slice

so this slice cannot contain newline in binary?

D3Hunter: seems ok to have raw newline in sql, like 
```sql
insert into t values('a
a');
```

D3Hunter (approved):

sync_diff_inspector/utils/utils.go
kennytm: @joechenrh I don't see why a `[]byte` cannot contain \x0D (\r) or \x0A (\n).

joechenrh: @kennytm  Sorry for the late reply, I missed the message.

I mean, we don't need to escape `\n`. Just like the example in the comment above, we can directly add raw newline in SQL:

```sql
mysql> insert into test.t1 values(1, "a\"
    "> \\1'`_%!");
Query OK, 1 row affected (0.01 sec)

mysql> select * from test.t1;
+------+------------+
| id   | col1       |
+------+------------+
|    1 | a"
\1'`_%! |
+------+------------+
1 row in set (0.00 sec)
```

PR #837, opened 2025-01-02
sync-diff-inspector: add `sql-hint-use-index` in config for automatically apply hint in check SQL

joechenrh:
### What problem does this PR solve?

Issue Number: close #836

Add hint `/*+ USE_INDEX(dbName.tblName, indexName) */` to force index scan in checksum SQL. Note this index hint is only applied on TiDB source.

How to use?

```
[data-sources]
[data-sources.source]
host = "127.0.0.1"
port = 4000
user = "root"
password = ""

# Automatically detects and applies existing indexes, only applies on TiDB source.
sql-hint-use-index = "auto"
```

### What is changed and how it works?

In each chunk, we will store the column names of the index used to split chunks. These column names will be used to match exact index name to build the index hint.

The reason we store **column names** rather than store index name directly is to avoid different index names between upstream and downstream.

For example:

- upstream has table `t1(id int, col1 int, unique index i1(id, col1))`
- downstream has table `t1(id int, col1 int, unique index i2(id, col1))`

### Check List 

Tests 

 - Unit test
 - Integration test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has exported function/method change
 - Has exported variable/fields change
 - Has interface methods change
 - Has persistent data change

Side effects

 - Possible performance regression
 - Increased code complexity
 - Breaking backward compatibility

Related changes

 - Need to cherry-pick to the release branch
 - Need to update the documentation
 - Need to be included in the release note

PR #857, opened 2025-05-13
sync-diff-inspector: Fix the indexColumns logic in random splitter to find matching index without sort

### LINKED ISSUE #858: sync-diff-inspector: sql-hint-use-index in random splitter can't find the right index in some cases
## Bug Report

1. What did you do?
The sql-hint-use-index may not pick the index in random splitter if the index cols order are not same as the original col order. The bug is in this [line](https://github.com/pingcap/tidb-tools/blob/ac97e96991cf66e7a81bc35e1238c084875f4e5f/sync_diff_inspector/splitter/random.go#L85C1-L88C4) where the cols in iFields is not sorted, but the MatchesIndex assumed it is.

E.g. for table
```
CREATE TABLE test1  (
  `id_second` INT NOT NULL,
   `id_first` INT NOT NULL,
    `v` VARCHAR(20)
  PRIMARY KEY (`id_first`,`id_second`)
);
```

The right index is picked

3. What did you see instead?
No index is picked.

5. which tool are you using?

gegao-db:

Issue Number: close #858

In random splitter when picking the index name for query hint, the iFields.MatchesIndex(index) sort the index cols before comparison, however we should use the exact index col order here, as the `fields` column list is already containing the ordered columns used for constructing the chunks.

This PR fixes this bug by not using MatchesIndex

### Check List 

Tests 

 - Unit test

### DIFF
--- a/sync_diff_inspector/splitter/random.go
+++ b/sync_diff_inspector/splitter/random.go
@@ -65,6 +65,8 @@ func NewRandomIteratorWithCheckpoint(ctx context.Context, progressID string, tab
 	// Below logic is modified from BucketIterator
 	// It's used to find the index which can match the split fields in RandomIterator.
 	iFields := &indexFields{cols: fields, tableInfo: table.Info}
+	// The cols needs to be sorted first for comparison in MatchesIndex method below
+	sortColsInPlace(iFields.cols)
 	var indices = dbutil.FindAllIndex(table.Info)
 NEXTINDEX:
 	for _, index := range indices {
--- a/sync_diff_inspector/splitter/random.go
+++ b/sync_diff_inspector/splitter/random.go
@@ -64,9 +64,7 @@ func NewRandomIteratorWithCheckpoint(ctx context.Context, progressID string, tab
 
 	// Below logic is modified from BucketIterator
 	// It's used to find the index which can match the split fields in RandomIterator.
-	iFields := &indexFields{cols: fields, tableInfo: table.Info}
-	// The cols needs to be sorted first for comparison in MatchesIndex method below
-	sortColsInPlace(iFields.cols)
+	fieldNames := utils.GetColumnNames(fields)
 	var indices = dbutil.FindAllIndex(table.Info)
 NEXTINDEX:
 	for _, index := range indices {
@@ -84,8 +82,7 @@ NEXTINDEX:
 			continue
 		}
 
-		if !iFields.MatchesIndex(index) {
-			// We are enforcing user configured "index-fields" settings.
+		if !utils.IsIndexMatchingColumns(index, fieldNames) {
 			continue
 		}
 
--- a/sync_diff_inspector/splitter/splitter_test.go
+++ b/sync_diff_inspector/splitter/splitter_test.go
@@ -997,6 +997,10 @@ func TestRandomSpliterHint(t *testing.T) {
 			[]model.CIStr{model.NewCIStr("a"), model.NewCIStr("b")},
 		},
 		{
+			"create table `test`.`test`(`a` int, `b` int, `c` int, primary key(`b`, `a`), unique key i1(`c`))",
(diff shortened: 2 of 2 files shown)


kennytm:
PTAL @joechenrh can you verify if this is compatible with #837

sync_diff_inspector/splitter/random.go
```diff
@@ -64,7 +64,7 @@ func NewRandomIteratorWithCheckpoint(ctx context.Context, progressID string, tab
 
 	// Below logic is modified from BucketIterator
 	// It's used to find the index which can match the split fields in RandomIterator.
-	iFields := &indexFields{cols: fields, tableInfo: table.Info}
+	fieldNames := utils.GetColumnNames(fields)
```
joechenrh, line 67: Maybe we should sort the `fields` before matching, just like what `BucketIterator` does.

The reason is that `fields` may come from `index-fields` in configuration file, which can be unordered. So we need to sort both `fields` and index columns before comparation.

gegao-db: The logic here is a bit different.
The `fields` here is already the one used by the random splitter below, so the order in `fields` actually matters, and we can't just find an index that doesn't match the order.

I've tried to see if I can copy the logic from bucket here, but in bucket splitter, it assumes the fields must have a matching index, while here in random splitter, it can use arbitrary fields even there is no corresponding index. 
So it is not easy to copy the exact logic.  

This does mean that it treat the `index-fields` in configuration file differently. Let me know if you have a better idea to fix this.

joechenrh: > This does mean that it treat the index-fields in configuration file differently.

I think we should have a unified standard, either sort fields in `RandomIterator` too, or remove sort logic in `BucketIterator`.

If we have an index `i1(c1, c2, c3)`, `index-fields = ["c1", "c3", "c2"]` can match this index in `BucketIterator` because fields are sorted in `indexFieldsFromConfigString`, while  `RandomIterator` can't after this PR.

gegao-db: Just to clarify 
> If we have an index i1(c1, c2, c3), index-fields = ["c1", "c3", "c2"] can match this index in BucketIterator because fields are sorted in indexFieldsFromConfigString, while RandomIterator can't after this PR.

This is the behavior before the PR as well. as random iterator respects the order in the user provided `index-fields`. That is why I wrote this way.

The difference is that:
Bucket iterator requires `index-fields` to match one of indexes on the table, even out of order, and it uses that index for the partitioning.
Random iterator doesn't require `index-fields` to match one of indexes, if it matches, it will use that index for hint, and if none matches, it will use the original order of cols for partitioning.

joechenrh: > as random iterator respects the order in the user provided `index-fields`

Yeah, I mean this behavior is quite weird. Since these fields are used to split chunks, order ain't matter. So if possible, could we change this in this PR as well.

gegao-db: I am a bit inclined to not make that change in this PR. 
The logic today is a bit intertwined, as the random splitter will first do one pass of index selection in `GetSplitFields`, and then do it a second time here for just picking the hint. 

And this would be a behavior change if user relies on the order of columns in the config.

Would like to keep this change simple to just pick the right index for hint purpose and doesn't change the chunking logic.

joechenrh: > And this would be a behavior change if user relies on the order of columns in the config.

Hmm, I think no user would rely on the order, as `index-fields` is described to only split the chunk in the document. The hint here is only to prevent from TiDB wrongly choosing table scan rather than index scan due to some statistics problem.

BTW, It's OK to keep it for now.

joechenrh (approved):

kennytm (approved):

PR #865, opened 2025-07-29
Use TLS by default

### LINKED ISSUE #864: sync-diff-inspector: doesn't use TLS if available
## Feature Request

Currently sync-diff-inspector only uses TLS when configured with a CA certificate.

The result is that it doesn't use TLS if there is no security section in the config.

This leads to failures if `require_secure_transport=ON` is used.

dveeden:

Issue Number: close #864 

This PR makes it:
- Use TLS if available. With fallback to non-TLS if it is not available.
- This will not do CA validation as there is no CA configured.
- This results that without TLS config `sync-diff-inspector` can connect to non-TLS servers and servers that use TLS, including `require_secure_transport=ON`. 

It is still recommended to configure a CA certificate and if needed a client certificate and key as the CA certificate can prevent against MitM attacks.

Alternatives
- We could add an option like `security.tlsmode` with values like `disabled`/`required`/`preferred`/`validate_ca`/`validate_identity` that follows the similarly named options in the MySQL Client.
- We could make an option that turns on TLS without fallback to non-TLS, but without the CA validation.

For the unittest
- Fixed the order of the actual and expected output for the lines that I touched
- Using `if testing.Testing()` to get a stable `TLSName` for testing.

### Check List 

Tests 

Related changes


dveeden:
Moved to https://github.com/pingcap/tiflow/pull/12268

PR #874, opened 2025-09-24
sync-diff-inspector:  refactor GetBucketsInfo from SHOW STATS_BUCKETS to SELECT query

### LINKED ISSUE #873: sync-diff-inspector should use `select mysql.stats_buckets`
## Feature Request

Sync-diff-inspector falls back to random bucket iterator when no bucket stats available, which results in a very slow sampling query to generate chunk boundaries for large tables.
Checked an error in sync-diff-inspector: it couldn’t find the bucket stats on the primary key id. However, TiDB actually does have stats_buckets for this column; they just weren’t loaded into the stats cache.

```
[2025/09/17 21:14:35.721 -07:00] [DEBUG] [common.go:528] [GetBucketsInfo] [sql="SHOW STATS_BUCKETS WHERE db_name= ? AND table_name= ?;"] [schema=xxx] [table=xxxx]
[2025/09/17 21:14:37.175 -07:00] [INFO] [tidb.go:61] ["failed to build bucket iterator, fall back to use random iterator"] [error="primary key on id in buckets info for xxx.xxxx not found"] 
``` 
We can support `select xxx from mysql.STATS_BUCKETS`

King-Dylan:

Issue Number: close #873

### Check List 

Tests 

 - Unit test
 - Integration test
 - No code

Code changes

 - Has persistent data change

Side effects

 - Increased code complexity

Related changes

### Release note

```release-note
Using a SELECT query instead of SHOW STATS_BUCKETS, so it works even when table statistics are not loaded into memory.
```


King-Dylan:
It may be necessary for TiDB to expose the column ID to merge this PR.

pkg/dbutil/common.go
```diff
@@ -508,79 +530,125 @@ type Bucket struct {
...
+		placeholders[i] = "?"
+		args[i] = id
+	}
+
+	query := fmt.Sprintf(`
+		SELECT is_index, hist_id, bucket_id, count, lower_bound, upper_bound
+		FROM mysql.stats_buckets
+		WHERE table_id IN (%s)
+		ORDER BY is_index, hist_id, bucket_id`,
+		strings.Join(placeholders, ","))
```
kennytm: why use 3 queries to get this rather than a single query?

```sql
select is_index, hist_id, bucket_id, count, lower_bound, upper_bound
from mysql.stats_buckets
where table_id in (
    select tidb_table_id from information_schema.tables where table_schema = ? and table_name = ?
    union all
    select tidb_partition_id from information_schema.partitions where table_schema = ? and table_name = ? 
) order by is_index, hist_id, bucket_id
```

King-Dylan: Thanks for the suggestion! Yes, we can switch to a single SQL that unions table and partition IDs and then queries mysql.stats_buckets—that’ll cut down on round trips. The only trade-off is that separate queries make error diagnosis clearer (whether ID retrieval failed vs. the stats_buckets read). I’ll update it.

pkg/dbutil/common.go
```diff
@@ -508,79 +508,91 @@ type Bucket struct {
 	UpperBound string
 }
 
-// GetBucketsInfo SHOW STATS_BUCKETS in TiDB.
+// QueryTableIDSubquery is the subquery to get table ID from information_schema.tables
+const QueryTableIDSubquery = "SELECT tidb_table_id FROM information_schema.tables WHERE table_schema = ? AND table_name = ?"
+
+// QueryPartitionIDSubquery is the subquery to get partition ID from information_schema.partitions
+const QueryPartitionIDSubquery = "SELECT tidb_partition_id FROM information_schema.partitions WHERE table_schema = ? AND table_name = ?"
```
kennytm: since we no longer have the separate `GetTableID` and `GetPartitionID` functions I don't see any reason why these 2 constants have to be separated out, even less reason making them public.

kennytm (commented):
rest lgtm

kennytm (approved):

PR #888, opened 2026-02-27
sync_diff_inspector: fix binsearch route table lookup when upstream is TiDB

### LINKED ISSUE #887: sync-diff-inspector: failed to generate fixed SQL when work source contains route rules
## Bug Report

1. What did you do?

This problem is raised from the customer. Below is a minimum config, the keypoint is that we use **TiDB with route rule as upstream**.

```toml

export-fix-sql = true

######################### Datasource config #########################
[data-sources]
[data-sources.mysql1]
    host = "localhost"
    port = xxx
    user = "xxx"
    route-rules = ["rule1"]

[data-sources.tidb0]
    host = "localhost"
    port = xxx
    user = "xxx"

########################### Routes ##############################
[routes]
schema-pattern = "src"
target-schema = "tgt"

######################### task config #########################
# Configures the tables of the target database that need to be compared.
[task]
    source-instances = ["mysql1"]
    target-instance = "tidb0"
    target-check-tables = ["tgt.*"]
```

sync-diff-inspector can output the fix SQL.

3. What did you see instead?

sync-diff-inspector failed with errors like 

```
 SELECT command denied to user 'xxx'@'127.0.0.1' for table 'tbl' error occurred in `tgt`.`tbl`
```

But the table we need to query is `src.tbl`, not `tgt.tbl`.

5. which tool are you using?

joechenrh:

Issue Number: close #887

## Root cause:
- In binsearch, the midpoint query (`GetApproximateMidBySize`) used `tableDiff.Schema/tableDiff.Table` (logical target table name).
- When upstream is TiDB with route rules, this logical name can map to a different physical upstream table name.
- As a result, **binsearch queried the wrong table (or a non-existent table) on upstream,** so midpoint splitting was incorrect and route-based compare could fail.

- Use physical source table name from route mapping during binsearch (`GetApproximateMidBySize`) instead of logical target table name.
- Unify source table lookup via `GetSourceTable(*splitter.RangeInfo)` in `Source` interface and implementations.
- Add nil-safe fallback to logical table name when route map lookup misses.
- Add integration tests with small datasets.

### Check List

Tests 

 - Unit test
 - Integration test

Manual test steps:
- `make failpoint-enable`
- `go test ./sync_diff_inspector/source -run 'TestTiDBSourceGetSourceTable|TestMySQLSourcesGetSourceTable' -count=1`
- `make sync_diff_inspector`
- `PATH=tests/_utils:bin:$PATH MYSQL_HOST=127.0.0.1 MYSQL_PORT=3306 sh tests/sync_diff_inspector/router/run.sh`
- `make failpoint-disable`

Code changes

Side effects

 - Increased code complexity


kennytm (approved):

Benjamin2037 (commented):
LGTM

PR #899, opened 2026-06-14
sync-diff-inspector: filter auto snapshot syncpoint by changefeed

### LINKED ISSUE #898: sync-diff-inspector auto snapshot can pick the wrong changefeed syncpoint
## Bug Report

1. What did you do?

Ran multi-tenant TiCDC with two different changefeeds streaming to the same downstream, then ran sync-diff-inspector with `snapshot = "auto"` for both upstream and downstream to compare one tenant/keyspace's tables.

In this setup, both changefeeds write syncpoint rows into the same downstream `tidb_cdc.syncpoint_v1` table. The current auto-snapshot query appears to select the latest syncpoint globally:

```sql
SELECT primary_ts, secondary_ts
FROM tidb_cdc.syncpoint_v1
ORDER BY primary_ts DESC
LIMIT 1;
```

Because the query is not correlated with the intended changefeed, sync-diff-inspector can choose the syncpoint row from another changefeed.

sync-diff-inspector should select the syncpoint row that belongs to the changefeed being compared, or otherwise require an unambiguous snapshot mapping. With the correct changefeed's syncpoint, the compared table should not be reported inconsistent.

3. What did you see instead?

sync-diff-inspector selected a syncpoint row from a different changefeed that was streaming to the same downstream. That produced a false-positive data mismatch because the downstream snapshot was slightly earlier than the matching changefeed's own downstream syncpoint.

Concrete evidence from the failed comparison:

- Source snapshot: `466946052587520000`
- Target snapshot selected by sync-diff-inspector: `466946054006767616`
- Matching changefeed's target snapshot should have been: `466946054033768448`
- Reported mismatch: `tpcc_ks2.warehouse`, `w_id = 2`, source `w_ytd = 1673384.64`, downstream `w_ytd = 1672005.02`

The selected target snapshot exactly matched another changefeed's syncpoint row, while the intended changefeed had a later syncpoint row for the same primary timestamp. This makes the mismatch a snapshot-selection/correlation problem rather than evidence that TiCDC lost the row update.

N/A. This report is about sync-diff-inspector auto snapshot selection.

5. which tool are you using?

sync-diff-inspector

sync-diff-inspector of master

Suggested follow-up:

- Make auto snapshot selection filter `tidb_cdc.syncpoint_v1` by the intended changefeed, if that information is available to sync-diff-inspector.
- Alternatively, let callers provide explicit per-changefeed `(primary_ts, secondary_ts)` snapshots for source and downstream.
- Or isolate syncpoint metadata so a global latest-row query cannot cross streams between independent changefeeds.

pingyu:

Issue Number: close #898

sync-diff-inspector currently resolves `snapshot = "auto"` by reading the latest row from `tidb_cdc.syncpoint_v1` globally. When multiple TiCDC changefeeds write syncpoint rows into the same downstream table, this can select another changefeed's syncpoint and produce false-positive data mismatches.

This PR adds an optional `syncpoint-changefeed` task config for sync-diff-inspector. When both source and target snapshots are set to `"auto"` and this option is configured, sync-diff-inspector filters `tidb_cdc.syncpoint_v1` by that changefeed before selecting the latest syncpoint row. If the option is omitted, the existing global latest-row behavior is preserved.

The change also updates the config template, includes the new field in the task config hash, and adds unit coverage for config parsing plus filtered syncpoint lookup behavior.

### Check List 

Tests 

 - Unit test (`go test ./sync_diff_inspector/config ./sync_diff_inspector/source`)

Code changes

### Release Note

```release-note
sync-diff-inspector supports filtering auto snapshot syncpoint lookup by TiCDC changefeed with the `syncpoint-changefeed` configuration.
```


pingyu:
Move to https://github.com/pingcap/tiflow/pull/12725

